PR #231, opened 2023-07-13
Use numpy for shuffling instead of torch

dirkgr:
Randomness with torch:
image

Randomness with numpy:
image


jpc:
Hi, I know I am a bit late to the party but the plots/images don't seem to load anymore. Reading about this in the ModernBERT paper got me really curious how bad the PyTorch RNG situation is.

Also, did anybody had a chance to measure the impact of this RNG change on model convergence?

dirkgr:
Twitter still has it: https://x.com/mechanicaldirk/status/1679348829945225216

PR #384, opened 2023-11-27
DeepSpeed

Muennighoff:
- Does not yet support checkpointing
- `configs/olmo-small-ablation-lumi-deepspeed.yaml` is the same as `configs/olmo-small-ablation-lumi.yaml` except for `deepspeed: true` & `init_device: cpu`
- `scripts/lumi/olmo-small-ablation-on-lumi-test.sh` is the same as `scripts/lumi/olmo-small-ablation-on-lumi-test-deepspeed.sh` except for `export CONFIG_PATH=configs/olmo-small-ablation-lumi-deepspeed.yaml`

### DIFF
--- a/olmo/train.py
+++ b/olmo/train.py
@@ -721,37 +721,37 @@ class Trainer:
                 process_group=self.fsdp_model.process_group,
             )
 
-        # Adjust the learning rate.
-        for group in self.optim.param_groups:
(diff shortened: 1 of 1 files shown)


Muennighoff:
For future reference, this PR works but we are not merging it as there's no speed-ups and it's not clean yet. Attached is a comparison of the throughput with this integraton vs FSDP :)

image

PR #393, opened 2023-12-11
Make hf_olmo device_map compatible

AkshitaB:


sarahwie:
This doesn't seem to be working as expected-- when I try running the following line
`> model = OLMoForCausalLM.from_pretrained("allenai/OLMo-7B", device_map="auto")`

I get the error
```
The model weights are not tied. Please use the `tie_weights` method before using the `infer_auto_device` function.
```

PR #414, opened 2024-01-24
Changes for 65B and 70B runs

dirkgr:


olmo/config.py
```diff
@@ -243,6 +242,14 @@ class ModelConfig(BaseConfig):
     The number of self-attention heads.
     """
 
+    n_kv_heads: Optional[int] = None
+    """
+    The number of heads to use for keys and values.
```
Muennighoff, line 247:
```suggestion
    The number of heads to use for keys and values. Defaults to `n_heads`.
```

olmo/config.py
```diff
@@ -428,6 +434,29 @@ class ModelConfig(BaseConfig):
...
+        if self.n_kv_heads is None:
+            self.n_kv_heads = self.n_heads
+
+    @classmethod
+    def update_legacy_settings(cls, config: D) -> D:
+        new_config = config.copy()
+        if om.is_dict(new_config):
+            assert isinstance(new_config, DictConfig)
+
+            if hasattr(new_config, "multi_query_attention"):
```
Muennighoff, line 447: I am probably missing sth but since ModelConfig is a dataclass won't it always have the attribute multi_query_attention? So maybe this should rather check if multi_query_attention is set to True and and n_kv_heads is set and error out in that case?

Muennighoff (approved):
The GQA looks good to me

olmo/config.py
```diff
@@ -309,8 +317,7 @@ class ModelConfig(BaseConfig):
 
     multi_query_attention: bool = False
```
epwalsh, line 317: Make this `Optional[bool]`, defaulting to `None`.

olmo/config.py
```diff
@@ -428,6 +435,29 @@ class ModelConfig(BaseConfig):
     See :data:`TrainConfig.precision` instead.
     """
 
+    def __post_init__(self):
+        if self.n_kv_heads is None:
+            self.n_kv_heads = self.n_heads
```
epwalsh, line 439: Then here we could do this:

```python
if self.multi_query_attention:
    self.n_kv_heads = 1
elif self.n_kv_heads is None:
    self.n_kv_heads = self.n_heads
```

olmo/config.py
```diff
@@ -428,6 +435,29 @@ class ModelConfig(BaseConfig):
...
+
+    @classmethod
+    def update_legacy_settings(cls, config: D) -> D:
```
epwalsh, line 442: Then this won't be needed.

epwalsh (commented):
Here's an alternative approach that doesn't involve implementing `update_legacy_settings()`.

olmo/config.py
```diff
@@ -428,6 +435,29 @@ class ModelConfig(BaseConfig):
...
+            if hasattr(new_config, "multi_query_attention"):
+                if hasattr(new_config, "n_kv_heads") and new_config.n_kv_heads is not None:
+                    raise OlmoConfigurationError("You can't specify both `multi_query_attention` and `n_kv_heads`. Specify only `n_kv_heads`.")
+                if new_config.multi_query_attention:
+                    new_config.n_kv_heads = 1
+                else:
+                    new_config.n_kv_heads = new_config.n_heads
+
+            if hasattr(new_config, "optimizer"):
+                new_config.optimizer = OptimizerConfig.update_legacy_settings(new_config.optimizer)
```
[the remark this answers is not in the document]
dirkgr: I am learning that `update_leacy_settings` doesn't work anyways with settings you specify on the command line.

olmo/config.py
dirkgr: Done, I just can't click "commit" here for some reason.

olmo/checkpoint.py
```diff
@@ -1376,12 +1374,12 @@ def save_checkpoint(
...
             )
 
-            # Save config.
-            self._save_config(checkpoint_dir, upload_to=upload_to)
-
             # Save metadata.
             self._save_metadata(checkpoint_dir, upload_to=upload_to)
 
+            # Save config.
+            self._save_config(checkpoint_dir, upload_to=upload_to)
```
dirkgr, line 1381: Maybe add a comment explaining why this is now last?

olmo/model.py
```diff
@@ -245,7 +260,7 @@ def __init__(self, config: ModelConfig, cache: BufferCache):
         self.config = config
         self.__cache = cache
         # Warm up cache.
-        self.get_rotary_embedding(config.max_sequence_length, _non_meta_init_device(config))
+        #  self.get_rotary_embedding(config.max_sequence_length, _non_meta_init_device(config))
```
dirkgr, line 263: We don't need this anymore?

scripts/beaker/mitchish70.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -ex
+
+CONFIG_PATH=configs/mitchish70-s3.yaml
+NUM_NODES=4
+SEED=3423
+INIT=fan_in
+RUN_NAME="fan-in-init-${SEED}"
```
dirkgr, line 9: I thought you said mitch init was best?

scripts/beaker/mitchish70.sh
```diff
@@ -0,0 +1,38 @@
...
+ARGS="--run_name=${RUN_NAME} --data.seed=6198 --seed=${SEED} --model.init_fn=${INIT} --model.init_std=0.006 --model.init_cutoff_factor=3 --device_train_microbatch_size=4 --model.flash_attention=true --fused_loss=true --evaluators=[] --stop_at=500 --wandb.group=mitchish70-ablate-init --save_interval_ephemeral=100"
```
dirkgr, line 10: Looks like this isn't the final config anyways.

dirkgr (commented):
I think there is some leftover debug commenting? Other than that, looks good.

olmo/model.py
epwalsh: Reverted.

olmo/checkpoint.py
epwalsh: c6c114d7

scripts/beaker/mitchish70.sh
epwalsh: Cleaned up in 92d2a08c.

PR #421, opened 2024-02-01
freeze official configs for reproductions

epwalsh:
Creates official copies of the configs we used to train the released models, with sensible names.

Still TODO:
- [x] Save 1B config to `configs/official/OLMo-1B.yaml`. @soldni 
- [x] Update paths in those configs to point to data on R2 instead of S3, once that's copied over. @2015aroras
- [x] Update README with pointers to these configs and examples on how to use them. @epwalsh


soldni:
Hey @epwalsh, added the 1B config and set the correct EOS token on both 1B and 7B. Didn't touch any data paths, lmk how you'd to handle it.

epwalsh:
Thanks @soldni, I'll update paths in both configs once the data is public on R2.

Issue #422, opened 2024-02-02
Missing Python Requirement


velocityCavalry:
### 📚 The doc issue

Hi there,

Thank you so much for open-sourcing OLMo! I tried to download the models from huggingface but encountered an attribute error as follows: 
```
>>> import hf_olmo
>>> from transformers import AutoModelForCausalLM, AutoTokenizer
>>> olmo = AutoModelForCausalLM.from_pretrained("allenai/OLMo-7B", cache_dir='olmo')
config.json: 100%|███████████████████████████████████████| 1.32k/1.32k [00:00<00:00, 504kB/s]
pytorch_model.bin: 100%|█████████████████████████████████| 27.6G/27.6G [01:53<00:00, 243MB/s]
/home/python3.9/site-packages/torch/_utils.py:831: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
  return self.fget.__get__(instance, owner)()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/lib/python3.9/site-packages/transformers/models/auto/auto_factory.py", line 566, in from_pretrained
    return model_class.from_pretrained(
  File "/home/lib/python3.9/site-packages/transformers/modeling_utils.py", line 3594, in from_pretrained
    model = cls(config, *model_args, **model_kwargs)
  File "/home/lib/python3.9/site-packages/hf_olmo/modeling_olmo.py", line 40, in __init__
    model_config = create_model_config_from_pretrained_config(config)
  File "/home/lib/python3.9/site-packages/hf_olmo/modeling_olmo.py", line 20, in create_model_config_from_pretrained_config
    for key in ModelConfig.__match_args__:
AttributeError: type object 'ModelConfig' has no attribute '__match_args__'
```
Upon further examination, I discovered that `__match_args__` seems to be integrated into python starting from 3.10. (https://peps.python.org/pep-0622/; https://peps.python.org/pep-0636/). I figure that if we add a python requirement to be >= 3.10, it will be better :)!

After updating the python version, the above code runs fine (with a minor warning):
```
>>> import hf_olmo
>>> from transformers import AutoModelForCausalLM, AutoTokenizer
>>> olmo = AutoModelForCausalLM.from_pretrained("allenai/OLMo-1B", cache_dir='olmo')
[...]

PR #424, opened 2024-02-02
Update README with Python Version Requirement

engichang1467:
This PR updates the README to include the minimum Python version requirement (>=3.10), addressing confusion and compatibility issues highlighted in issue #422 

### Changes include:

- A new section at the beginning of the Installation instructions, guiding users to ensure they have Python 3.10 or later before proceeding with the installation process.
- Instructions for checking the current Python version.
- A note encouraging users to update their Python version if it is below 3.10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,14 @@ It is built by scientists, for scientists.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


epwalsh:
As of #426 this shouldn't be necessary. Thank you though!

PR #427, opened 2024-02-03
Update README.md with paper references 

ViswanathaReddyGajjala:
Added paper references.


bhAvA2004:
good......

epwalsh:
Thanks @ViswanathaReddyGajjala, I've added a slightly different Bibtex entry in #421 and excluded the Dolma one since that can found in the Dolma repo.

PR #442, opened 2024-02-09
[HF] Add input embedding argument to HF model

2015aroras:
The trainers in HF seem to expect models to accept input embeddings directly as an alternative to input ids. This expectation caused [this HF issue](https://huggingface.co/allenai/OLMo-7B/discussions/7).

This PR adds an input embeddings argument to OLMo, and bypasses the input embedding layer when input embeddings are provided.

### DIFF
--- a/olmo/model.py
+++ b/olmo/model.py
@@ -1137,6 +1137,7 @@ class Olmo(nn.Module):
     def forward(
         self,
         input_ids: torch.LongTensor,
+        input_embeddings: Optional[torch.FloatTensor] = None,
         attention_mask: Optional[torch.Tensor] = None,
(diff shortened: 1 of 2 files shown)


olmo/model.py
```diff
@@ -1137,6 +1137,7 @@ def get_alibi_attention_bias(self, seq_len: int, device: torch.device) -> torch.
     def forward(
         self,
         input_ids: torch.LongTensor,
```
2015aroras, line 1139: Ideally I would make `input_ids` optional, but this would not be a fully backwards compatible change.

epwalsh (approved):
LGTM. Can you add an entry to the CHANGELOG?

PR #443, opened 2024-02-10
GQA into Mitchich65

dirkgr:
TODO
 - [x] Doesn't run on AMD like this


tests/model_test.py
```diff
@@ -232,6 +234,76 @@ def test_forward(
...
+        train_config.model.multi_query_attention = multi_query_attention
+        train_config.model.init_device = "cuda"
+        return Olmo(train_config.model).eval()
+
+    input1 = tokenizer.encode("As a large language model, I don’t have personal opinions.")
+    input2 = tokenizer.encode("But I can share some interesting facts!")
+    batch_inputs = DataCollator.from_train_config(train_config)(
+        [  # type: ignore
+            {"input_ids": input1, "attention_mask": [1.0] * len(input1)},
+            {"input_ids": input2, "attention_mask": [1.0] * len(input2)},
```
epwalsh, line 278: I don't think we're actually using `flash-attn` here since we're passing an attention mask.

epwalsh: I would just remove the mask here.

dirkgr: Haha, yes, I noticed all those same issues and fixed them before I read your comment here 😮‍💨.

scripts/beaker/mitchish70.sh
```diff
@@ -0,0 +1,33 @@
...
+NUM_NODES=4
+ARGS='--device_train_microbatch_size=4'
+
+gantry run \
+  --allow-dirty \
+  --workspace ai2/llm-testing \
+  --task-name mitchish65 \
+  --description mitchish65 \
+  --priority high \
+  --beaker-image olmo-torch2-gantry \
```
dirkgr, line 15: I think, at the moment, this will run on torch 2.0.

dirkgr:
@epwalsh, the fused CE loss, will it work on LUMI?

epwalsh:
@dirkgr I did add a config option to toggle the fused loss function in [54919e0](https://github.com/allenai/OLMo/pull/443/commits/54919e020e78d70b682b9295236b784eef2d0ed5). Defaults to false to be safe so will never run on LUMI.

olmo/config.py
```diff
@@ -1047,6 +1049,11 @@ class TrainConfig(BaseConfig):
     The activation checkpointing strategy to use.
     """
 
+    fused_loss: Optional[bool] = None
+    """
+    Whether to use the fused CE loss function from `flash-attn`.
```
dirkgr, line 1054: What's the default?

olmo/eval/evaluator.py
```diff
@@ -56,8 +56,8 @@ def compute_metrics(self) -> Dict[str, float]:
                     # get to this one within the current evaluation loop.
                     continue
                 else:
-                    out[f"eval/{label}/CrossEntropyLoss"] = loss.item()
-                    out[f"eval/{label}/Perplexity"] = torch.exp(loss).item()
+                    out[f"eval/ppl/{label}/CrossEntropyLoss"] = loss.item()
+                    out[f"eval/ppl/{label}/Perplexity"] = torch.exp(loss).item()
```
dirkgr, line 60: This will make comparisons impossible?

scripts/train.py
```diff
@@ -241,12 +242,18 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
 
 
 if __name__ == "__main__":
-    mp.set_start_method("spawn")
+    try:
+        mp.set_start_method("spawn", force=True)
+    except RuntimeError as e:
+        print(f"failed to set multiprocessing start method: {e}")
```
dirkgr, line 248: I did not know this could fail?

dirkgr (commented):
I cannot approve this because I created the OR. I am OK with this, with two concerns in comments.

scripts/train.py
epwalsh: Yea I saw this happen a couple times, not sure why exactly.

olmo/config.py
epwalsh: Defaults to `False`.

PR #449, opened 2024-02-13
Make clear that `output_hidden_states` and `output_attentions` aren't implemented 

natolambert:
Closed #447

PR #451, opened 2024-02-13
add capability to output intermediate hidden states

sarahwie:
Add a functionality that HF model forward calls have. 
A very similar update can be done to add the `output_attentions` functionality, though I haven't written it yet.

Note that HF code "appends" hidden states to a tuple object, but I'm using a list as I think it's better practice. Hence the type casting in the function output to return a tuple, in order to match expected output in HF's convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed default value of `--tokenizer` argument to `scripts/prepare_tulu_data.py` to be an absolute path, not relative path, the script can be run from other directories.
 - Added the option to directly pass input embeddings to `OLMo` and `OLMoForCausalLM`.
 - Added support for Python 3.8.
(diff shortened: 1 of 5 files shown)


natolambert:
Have you tested this @sarahwie ? Looks close to what I would expect but I don't use the OLMo repo much :)

AkshitaB:
Hi @sarahwie this is great! Do you mind also adding the use of the two flags to the HF wrapper here: https://github.com/allenai/OLMo/blob/main/hf_olmo/modeling_olmo.py#L48

AkshitaB (approved):

sarahwie:
I've tested this locally but haven't re-built the pip package, let me try that to make sure it still works as expected @natolambert

sarahwie:
Temporarily added code to throw an error if `output_attentions=True` (#449) since that functionality hasn't been coded yet as @natolambert pointed out.

 Tested this, and everything is worked as expected. Should I merge?

natolambert:
@sarahwie probably, but I'm not a core contributor. If you changed a bunch more, you can ask @AkshitaB to review the recent changes.

PR #456, opened 2024-02-18
get_fs_local_rank

hxdtest:


2015aroras:
I don't think this is the right approach to this problem. Our code (arbitrarily) assumes that unless `FS_LOCAL_RANK` is set, each node has a separate file system. I don't think assuming that each node has the same file system is a better assumption to have. The best behavior might be to raise an error and tell the user to explicitly set `FS_LOCAL_RANK`, so that no assumption is made.

PR #465, opened 2024-02-23
Uses flash-attn's "fused" CE loss function, when available

epwalsh:
I think they call this "fused" because it computes z-loss at the same time. As a result we see a bigger speedup when z-loss is enabled. See benchmarks below.

## Benchmarks

I didn't have the resources to run a full scale 70B test, so instead I ran a test with a 2-layer version of our 70B on 4 GPUs. Here is the speedup over what's on `main`.

- Z-loss disabled: 1.2% speedup
- Z-loss enabled: 3.5% speedup

### DIFF
--- a/olmo/train.py
+++ b/olmo/train.py
@@ -125,8 +125,10 @@ class Trainer:
                 cross_entropy_loss,
             )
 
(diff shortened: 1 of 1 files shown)


dirkgr (approved):
This is fine, might as well, but after 80 layers this 1.5% will disappear completely.

Issue #457, opened 2024-02-19
Olmo / OLMo consistency


Muennighoff:
### 🐛 Describe the bug

In the HF code we use OLMo but in training it's Olmo - This creates some inconsistencies when importing from the training modeling file ---- I think we should settle on one for both - wdyt? Probably OLMo as it's the actual name; happy to open a pr
(filing as a bug as e.g. `_no_split_modules = ["OLMoBlock"]` doesn't work i think)

cc @AkshitaB @dirkgr 

### Versions

latest

dirkgr:
Please file a PR! In the code, they should all be "OlmoSomething".

Muennighoff:
> Please file a PR! In the code, they should all be "OlmoSomething".

Sure so I'll rename everything in here https://github.com/allenai/OLMo/blob/main/hf_olmo/modeling_olmo.py to `Olmo`?

dirkgr:
👍🏻

Muennighoff:
Sounds good - will just wait for @AkshitaB to give her okay, as I will need to change a bunch of things including the config of the model on the hub https://huggingface.co/allenai/OLMo-7B/blob/main/config.json

AkshitaB:
Huggingface recommended the convention to be OLMoXYZ for the naming of the model. Also, I will recommend not making changes to the model on the hub, as it's not just 1 model, it's ALL the 1000+ checkpoints for all 3 models.

dirkgr:
Ok, then we'll go with Huggingface's suggestion. That means we rename everything to `OLMo*`, right?

Muennighoff:
Fixed it here: https://github.com/allenai/OLMo/pull/466

PR #466, opened 2024-02-25
Rename

Muennighoff:
Fixes https://github.com/allenai/OLMo/issues/457

As this is a breaking change, it probably makes sense to release a new version with it


dirkgr:
@epwalsh , @AkshitaB , @2015aroras , how much stuff will break if we merge this? Is it risky?

dirkgr:
I agree with you on the version, so I bumped the version number in this PR.

epwalsh:
I don't think trainer code or checkpoints will break, so I'm okay with it. Better to make this change now anyway before the next big run.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added the option to directly pass input embeddings to `OLMo` and `OLMoForCausalLM`.
 - Added support for Python 3.8.
 - Added code to throw an error if `output_attentions` is set to `True` in forward call to `OLMoForCausalLM`. This functionality hasn't been implemented yet.
+- Rename `Olmo` to `OLMo` everywhere in the codebase
```
2015aroras, line 18: Please move this to Unreleased or v0.3.0 or similar

2015aroras (commented):
The only thing I can foresee breaking is trying to run OLMo via HF. I think that's probably inevitable for a small period of time.

dirkgr (approved):

Issue #460, opened 2024-02-20
Does not support flash attention 2.0 on transformers.AutoModelForCausalLM.from_pretrained 


KaifAhmad1:
### 🚀 The feature, motivation and pitch

I am using Olmo 7B for RAG for efficient inference on low GPU resources but does not support flash attention 2.0 
Here is the code 
``` Python 
model = transformers.AutoModelForCausalLM.from_pretrained(
    model_id,
    config=model_config,
    device_map='auto',
    use_flash_attention_2="flash_attention_2",
    use_auth_token=hf_auth,
    quantization_config=bnb_config,
    low_cpu_mem_usage=True
)
``` 

```
The model was loaded with use_flash_attention_2=True, which is deprecated and may be removed in a future release. Please use `attn_implementation="flash_attention_2"` instead.
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[<ipython-input-17-96fef6444c74>](https://localhost:8080/#) in <cell line: 1>()
----> 1 model = transformers.AutoModelForCausalLM.from_pretrained(
      2     model_id,
      3     config=model_config,
      4     device_map='auto',
      5     use_flash_attention_2="flash_attention_2",

3 frames
[/usr/local/lib/python3.10/dist-packages/transformers/modeling_utils.py](https://localhost:8080/#) in _check_and_enable_flash_attn_2(cls, config, torch_dtype, device_map, check_device_map, hard_check_only)
   1465         """
   1466         if not cls._supports_flash_attn_2:
-> 1467             raise ValueError(
   1468                 f"{cls.__name__} does not support Flash Attention 2.0 yet. Please request to add support where"
   1469                 f" the model is hosted, on its model hub page: [https://huggingface.co/{config._name_or_path}/discussions/new](https://huggingface.co/%7Bconfig._name_or_path%7D/discussions/new)"

ValueError: OLMoForCausalLM does not support Flash Attention 2.0 yet. Please request to add support where the model is hosted, on its model hub page: https://huggingface.co/allenai/OLMo-7B/discussions/new or in the Transformers GitHub repo: https://github.com/huggingface/transformers/issues/new
```

dumitrac:
I apologize for our delay in response. In order to help surface current, unresolved issues, we are closing tickets prior to February 29.
Please reopen your ticket if you are continuing to experience this issue.
Please also see #557 - it might answer this question too.
Thank you!

PR #471, opened 2024-02-29
hf_olmo: support flash attn 2

wade3han:
https://github.com/allenai/OLMo/issues/460, tested with a simple snippet as below:

```
import transformers, torch

model = transformers.AutoModelForCausalLM.from_pretrained("allenai/OLMo-7B-Instruct", use_flash_attention_2="flash_attention_2", trust_remote_code=True).cuda()
tokenizer = transformers.AutoTokenizer.from_pretrained("allenai/OLMo-7B-Instruct", trust_remote_code=True)

print(tokenizer.decode(model.generate(torch.tensor(tokenizer.encode("Hello World! My name is")).unsqueeze(0).cuda())[0]))
# Hello World! My name is Emily and I am a second year student at the University of California,
```

### DIFF
--- a/hf_olmo/modeling_olmo.py
+++ b/hf_olmo/modeling_olmo.py
@@ -33,6 +33,7 @@ class OLMoForCausalLM(PreTrainedModel):
     config_class = OLMoConfig
     base_model_prefix = "model"
     _no_split_modules = ["OLMoBlock"]
+    _supports_flash_attn_2 = True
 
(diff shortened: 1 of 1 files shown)


epwalsh:
Someone familiar with `transformers` internals should review this (maybe @AkshitaB). I'm not sure what `transformers` does with this, but I'd be very cautious if they're monkey-patching our attention mechanism since flash-attn expends a different input shape (the head and sequence dimensions are flipped compared to what we normally do).

PR #476, opened 2024-03-01
Add support for PyTorch v2.2

epwalsh:
The main blocker was making sure our hacked-together `LocalShardedCheckpointer` class was updated to account for the recent changes to FSDP internals. I've verified that this checkpointer still works correctly. Here is a restart that reproduces the original loss curve perfectly:

image

See https://wandb.ai/ai2-llm/torch-22-testing?workspace=user-epwalsh for full logs.


dirkgr (approved):
That's all? 🤪

PR #480, opened 2024-03-04
Firehose Logging

dirkgr:


olmo/util.py
```diff
@@ -59,6 +59,7 @@ def __repr__(self) -> str:
 class LogFilterType(StrEnum):
     rank0_only = "rank0_only"
     local_rank0_only = "local_rank0_only"
+    firehose = "firehose"
```
epwalsh, line 62: I don't have a strong opinion here but maybe "all_ranks" instead of "firehose" to be more consistent with the existing names?

dirkgr: Done in [dd454bd](https://github.com/allenai/OLMo/pull/480/commits/dd454bdaced49f9809162a1f7c7d667c45a8091f).

2015aroras (approved):

dirkgr:
Tests fail because of a version conflict with cached_path. Cached_path can't be updated because of the script that parses the changelog. Looks like AGI has to wait another day.

epwalsh:
Oyvind has a fix in this PR: https://github.com/allenai/OLMo/pull/484

dirkgr:
> Oyvind has a fix in this PR: https://github.com/allenai/OLMo/pull/484

This is true, but the underlying issue is an unnecessary pin in cached_path.

PR #483, opened 2024-03-04
[Storage Cleaner] Unsharding improvements

2015aroras:
This is the first of a few PRs with refactored versions of changes I have been using locally for checkpoint management. This PR focusses on my changes to the storage cleaner's unsharding functionality:

1. Add the option to unshard just 1 checkpoint within a run. This is helpful if you want to unshard checkpoints in parallel using something like GNU parallel.
2. Add the option to delete a sharded checkpoint after it has been unsharded. This stops local file storage from blowing up.
3. Add support for unsharding checkpoints whose configs are not compatible with the current schema in `TrainConfig`. This has been done in a hacky way (passing an empty `TrainConfig` to the checkpointer), relying on the fact that unsharding doesn't use the config at all.

### DIFF
--- a/scripts/storage_cleaner.py
+++ b/scripts/storage_cleaner.py
@@ -15,6 +15,7 @@ from urllib.parse import urlparse
 import boto3.session
 import botocore.exceptions as boto_exceptions
 import google.cloud.storage as gcs
+import omegaconf
 import torch
 import wandb
 from boto3.s3.transfer import TransferConfig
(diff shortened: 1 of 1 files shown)


scripts/storage_cleaner.py
```diff
@@ -781,17 +788,31 @@ def _get_checkpoint_number(checkpoint_dir: str) -> int:
...
         sharded_checkpoint_directories = (
             [latest_checkpoint_directory] if latest_checkpoint_directory is not None else []
         )
+    elif checkpoint_num is not None:
+        sharded_checkpoint_directories = [
+            sharded_checkpoint_dir
+            for sharded_checkpoint_dir in sharded_checkpoint_directories
+            if _get_checkpoint_number(sharded_checkpoint_dir) == checkpoint_num
+        ]
+        assert len(sharded_checkpoint_directories) <= 1
```
epwalsh, line 815: Will this script warn anywhere if there are no matching checkpoints here?

epwalsh (approved):
LGTM, just one question

scripts/storage_cleaner.py
2015aroras: It will say (as an info log) there are no matching directories and then will exit gracefully.

PR #484, opened 2024-03-05
Add MMLU prompt variants

OyvindTafjord:
With @yulinggu-cs, added a version of the MMLU datasets with multiple prompt variants, to make the datasets effectively 7x larger. These are named `mmlu_stem_var` etc (add `_var` suffix).

The prompts are called: `[None, "inst", "inst+1", "inst+2", "inst+3", "inst+4", "inst+5"]`

where "inst+2" means add instruction line followed by 2-shot example. We also added logging of first few examples for each downstream task (can remove this if it's annoying, but it's a useful sanity check), here's an example for the "inst+2" prompt:
```
[2024-03-04 19:06:09] INFO     [olmo.eval.downstream:224, rank=0] Sample doc from (hails/mmlu_no_train, high_school_government_and_politics, inst+2):
doc_text: The following are multiple choice questions (with answers) about high school government and politics:

Question: Uncertainty over the limits to presidential power is caused primarily by the fact that
Answer: the constitutional definition of those powers is broad and unspecific

Question: The term "budget deficit" refers to the
Answer: amount the government spends in excess of its revenues

Question: Which of the following accurately describes congressional committees? I. The committee chairpersons always belong to the majority party. II. Seats on each committee are divided between the two major parties in exact proportion to the parties' representation in Congress. III. They recommend whether Congress should pass various pieces of legislation, and those recommendations are always approved by the full congressional body. IV. When a committee vote results in a tie, the vice president casts the tie-breaking vote.
Answer:
continuations: [' I only', ' II only', ' I and III only', ' II and III only']
```

Things to note:
   * There was a bug when an instance goes beyond max context length, as the computation in `update` uses `batch["ctx_len"]` which might be larger than 2048. I tried to fix this in https://github.com/allenai/OLMo/commit/f64b9ce1dffe6f5c61ad582f8c731a85c47f4c17, but it's a bit iffy (the context length is separately processed in `prep_examples` and in `collate_fn` which could be confusing
   * Logging the instances we noticed a tiny issue with the SciQ formatting, fixed [here](https://github.com/allenai/OLMo/commit/dff32c6882ccbc53a47a125f31efb6b0201db2b0#diff-cf39bf0a6cccecf47af2cebf6e33f2e452d7c66857aee654e5c0736b3f7aed4aL673)
   * As mentioned, we the first 5 instances in each dataset. If this happens in some sort of parallel processing, the method of doing `self.log_instances -= 1` might not be very robust? If this is too noisy in the logging, we can comment these lines out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added `output_hidden_states` argument and associated functionality to `OLMo` and `OLMoForCausalLM` to return model intermediate hidden states.
-- Added MMLU downstream evaluation tasks.
+- Added MMLU downstream evaluation tasks, with prompt variations.
 - Added support for PyTorch v2.2.
 
 ## [v0.2.4](https://github.com/allenai/OLMo/releases/tag/v0.2.4) - 2024-02-02
--- a/olmo/eval/downstream.py
+++ b/olmo/eval/downstream.py
@@ -1118,8 +1118,8 @@ class MMLU(ICLMultiChoiceTaskDataset):
             if num_shots:
                 dev_set = self.dev_set.get(doc.get('subject'), [])
                 for dev_doc in dev_set[:int(num_shots[0])]:
-                    answer = dev_doc["choices"][doc["answer"]]
-                    prefix += "Question: " + doc["question"] + "\nAnswer: " + answer + "\n\n"
+                    answer = dev_doc["choices"][dev_doc["answer"]]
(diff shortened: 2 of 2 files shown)


OyvindTafjord:
Huh, interesting, a test is failing with an error I ran into in the oe-eval-internal yesterday as well, regarding downloading datasets from HF hub. This will probably affect any OLMo run regardless of this PR. After some mad testing of python dependencies, I eventually found that [pinning](https://github.com/allenai/oe-eval-internal/commit/e0f261e62eda2ac1c2e8e5f96b181fcc38d44ae1) `fsspec==2023.5.0` resolved the issue. I assume this will be fixed quickly, but I didn't see any chatter about it when I did a quick search yesterday.

PR #486, opened 2024-03-06
Refactor monkey patch for loading in legacy unsharding

2015aroras:
This PR creates a context manager for the monkey patch for loading legacy sharded checkpoints. This makes turning the monkey patch on & off much cleaner (before we were not turning it off!). The main reason for this change is to reduce the size of my upcoming PR where I will change legacy unsharding to use processes and shared memory.


epwalsh (approved):
LGTM

PR #488, opened 2024-03-06
[Storage Cleaner] Speed up unsharding of some legacy checkpoints

2015aroras:
This PR changes the unsharding mechanism of legacy checkpoints to use processes and shared memory instead of threads. In one case where the world size was 1024, this implementation brought the unsharding time down from 6 hours to 30 minutes. This implementation is slower than the old one at smaller scales, but that is ok.

An option was to keep the old mechanism around in code too, but since we are trying to get rid of legacy sharded checkpoints it doesn't seem worth to keep that code around.

### DIFF
--- a/olmo/checkpoint.py
+++ b/olmo/checkpoint.py
@@ -962,13 +962,14 @@ class TorchLegacyShardedCheckpointer(Checkpointer):
         if rank_size == 0:
             return
 
-        temp: np.ndarray = torch.zeros(rank_size, dtype=shard0_md.tensor_properties.dtype).numpy()
-        numpy_type = temp.dtype
+        assert shard0_md.tensor_properties.dtype == torch.float32, "Expected sharded tensor to be fp32"
+        numpy_type = np.float32
 
         sharded_memory_name = "-".join(key + (str(rank),))
 
-        shm = shared_memory.SharedMemory(create=True, size=temp.nbytes, name=sharded_memory_name)
-        del temp
+        shm = shared_memory.SharedMemory(
+            create=True, size=rank_size * np.dtype(numpy_type).itemsize, name=sharded_memory_name
+        )
         np_arr = np.ndarray((rank_size,), dtype=numpy_type, buffer=shm.buf)
 
         for local_shard in sharded_tensor.local_shards():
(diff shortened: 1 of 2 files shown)


olmo/checkpoint.py
```diff
@@ -913,6 +916,162 @@ def unshard_checkpoint(
...
+        shard_placement, rank_sizes = self._get_shard_placement_and_rank_sizes(
+            shard0_md.shards_metadata, world_size
+        )
+
+        rank_size = rank_sizes[rank]
+        assert rank_size >= 0
+        if rank_size == 0:
+            return
+
+        temp: np.ndarray = torch.zeros(rank_size, dtype=shard0_md.tensor_properties.dtype).numpy()
```
epwalsh, line 965: What's the purpose of this temp array? If it's just to get the number of bytes, you should be able to infer that from the data type and size.

olmo/checkpoint.py
```diff
@@ -913,6 +916,162 @@ def unshard_checkpoint(
...
+
+        def shard_size(shard_md):
+            return reduce((lambda x, y: x * y), shard_md.shard_sizes)  # type: ignore[attr-defined]
+
+        shard_placement, rank_sizes = self._get_shard_placement_and_rank_sizes(
+            shard0_md.shards_metadata, world_size
+        )
+
+        temp: np.ndarray = torch.zeros(1, dtype=shard0_md.tensor_properties.dtype).numpy()
+        numpy_type = temp.dtype
```
epwalsh, line 1036: Similar question here, but looks like it's just to get the data type? It's probably reasonable to assume FP32.

epwalsh (commented):
Very cool! Just a couple questions

olmo/checkpoint.py
2015aroras: Just type and number of bytes. I've changed the code to assume fp32. I already know they are not bf16 at least because `.numpy()` fails for bf16.

olmo/checkpoint.py
2015aroras: https://github.com/allenai/OLMo/pull/488/commits/c58f4b4c9bafbce91315df9d9da240ff37cb13a9 Changed to assume fp32

olmo/checkpoint.py
epwalsh: yea when we train in bf16 (or fp16), the main copy of model weights is always fp32

epwalsh (approved):
LGTM

PR #489, opened 2024-03-07
Add option for QKV clipping

epwalsh:
Adds a config field for enabling QKV clipping: `clip_qkv: Optional[float]`.


AkshitaB (approved):
Looks fine. Should we add a check for the clipping value to be positive, since we negate it for getting the `min` value?

PR #493, opened 2024-03-08
[Storage Cleaner] Improvements to moving runs

2015aroras:
This is the first of 3 PRs that make changes to the move mechanism of the storage cleaner. The other PRs will:
1. Add the option to move just 1 entry from a run, making the script more parallelizable.
2. Migrate off `TrainConfig` to support legacy checkpoints.

This PR makes the following changes:

- Adds most of the functionality for moving non-checkpoint entries when `append_wandb_path` is set. This will make it possible to upload `train_data` to all folders that need it.
- Lets the storage adapter choose what happens when moving an entry to a destination where the entry already exists. This change is not necessary but is convenient.


epwalsh (approved):
LGTM

PR #494, opened 2024-03-09
[Storage Cleaner] Move 1 entry within a run directory

2015aroras:
This PR adds the option to move a single entry (file or directory) from a run to another location. This can be run in parallel for all entries run to get higher throughput.


scripts/storage_cleaner.py
```diff
@@ -1440,6 +1471,11 @@ def _add_move_subparser(subparsers: _SubParsersAction):
         action="store_true",
         help="If set, the wandb path for the run is found and appended to the destination dir. If the run is being stored as an archive file, wandb id is first removed from the wandb path and used as the filename.",
     )
+    move_parser.add_argument(
+        "--entry",
+        default=None,
+        help="If provided, moving is restricted to this entry of the run.",
```
epwalsh, line 1477: Maybe add a few words here to describe how an `--entry` should be specified. For example it looks it should be given as a filename without any other path components?

scripts/storage_cleaner.py
```diff
@@ -1211,12 +1229,22 @@ def _get_src_dest_pairs_for_copy(
         for wandb_path in set(checkpoint_to_wandb_path.values())
     ]
 
+    if config.entry is not None:
+        src_dest_pairs = [
+            src_dest_pair
+            for src_dest_pair in src_dest_pairs
+            if Path(src_dest_pair[0]).name == config.entry.rstrip("/")
```
epwalsh, line 1236: This seems a little bit brittle, what about:

```suggestion
            if str(Path(src_dest_pair[0])).endswith(str(Path(config.entry)))
```

Do we warn anywhere if this list comes up empty?

epwalsh (approved):
LGTM, just minor comments

scripts/storage_cleaner.py
2015aroras: https://github.com/allenai/OLMo/pull/494/commits/6f56700a78eec2667792541e8147b868d20a0419 I found `Path.match` and it seems more robust (it won't match `test123` when `config.entry` is `123`, for example).

https://github.com/allenai/OLMo/pull/494/commits/b9268411f28e8f9122e5bd452e946e9be250f90f Made empty list an error.

scripts/storage_cleaner.py
2015aroras: https://github.com/allenai/OLMo/pull/494/commits/90648bddb3e2ace234658d60c10c7c1928ba5b11

Issue #491, opened 2024-03-08
Shape mismatch error when resizing token embeddings in OLMo modeling code


djliden:
### 🐛 Describe the bug

I encountered a shape mismatch error when fine-tuning OLMo-1B using the Hugging Face Trainer after resizing the token embeddings. The error occurs due to an inconsistency between the `embedding_size` in the model configuration and the actual size of the token embeddings after resizing.

## Steps to Reproduce:
(See below for reproducible example)

1. Load an OLMo model and tokenizer with transformers library.
2. Add special tokens to the tokenizer.
3. Resize the token embeddings using `model.resize_token_embeddings(len(tokenizer))`.
4. Fine-tune the model using the Hugging Face Trainer.

Error Message: `RuntimeError: shape '[-1, 50304]' is invalid for input of size 102772320`

## Explanation:
The OLMo modeling code assumes that the `embedding_size` in the configuration (`self.config.embedding_size`) will always match the actual size of the token embeddings. However, when the token embeddings are resized using `resize_token_embeddings`, the `embedding_size` in the configuration remains unchanged, while the actual size of the token embeddings is updated. Note that `resize_token_embeddings` *does* update `self.config.vocab_size` but not `self.config.embedding_size` ([here](https://github.com/huggingface/transformers/blob/092f1fdaa4224fdd88c616dc9678e6fcb37bfffd/src/transformers/modeling_utils.py#L1761) is the implementation in the transformers library).

In the forward pass, the code reshapes the `shift_logits` tensor using `self.config.embedding_size`:

`shift_logits = shift_logits.view(-1, self.config.embedding_size)` ([link](https://github.com/allenai/OLMo/blob/752353bf5ad7ff5bdac40812dd6ad9a3ac55e001/hf_olmo/modeling_olmo.py#L90))

This leads to a shape mismatch error because the second dimension of `shift_logits` is expected to have a size of `self.config.embedding_size` (50304 in this case), but the actual size is the updated vocabulary size after resizing the token embeddings (e.g., 50282 after adding special tokens; because vocab size < embedding size, resizing according to the vocab size shrinks the embedding size).

## Proposed fixes
I see a few different options for addressing this, and am happy to contribute.
- The OLMo modeling code could be updated to use the actual size of the token embeddings (`self.transformer.wte.weight.shape[0]`) instead of relying on the `embedding_size` value from the configuration when reshaping `shift_logits`. This will ensure that the code works correctly even when the token embeddings are resized dynamically.
[...]

PR #501, opened 2024-03-13
Add custom `resize_token_embeddings` method to `OLMoForCausalLM` (#491)

djliden:
This PR addresses #491 by adding a custom `resize_token_embeddings` method to override the corresponding method from the `PreTrainedModel` base class. The revised method resizes `self.config.embedding_size` and `self.model.config.embedding_size` rather than resizing the `vocab_size` as done by the base class's method.

Here's the result with the new method; this resulted in an error previously (see #491).

```python
!pip install -e .
!pip install accelerate
dbutils.library.restartPython()

from transformers import AutoTokenizer, AutoModelForCausalLM
import torch
import hf_olmo

model_ckpt = "allenai/OLMo-1B"
tokenizer = AutoTokenizer.from_pretrained(model_ckpt)
model = AutoModelForCausalLM.from_pretrained(
    model_ckpt,
    torch_dtype=torch.bfloat16,
    device_map="auto",
    trust_remote_code=True,
)

print("Before resizing:")
print(f"Vocabulary size: {model.config.vocab_size}")
print(f"Embedding size: {model.config.embedding_size}")
print(f"Model embedding dimension: {model.get_input_embeddings().embedding_dim}")

special_tokens = ["<|im_start|>", "<|im_end|>"]
tokenizer.add_special_tokens({"additional_special_tokens": special_tokens})
model.resize_token_embeddings(len(tokenizer))

print("\nAfter resizing:")
print(f"Vocabulary size: {model.config.vocab_size}")
print(f"Embedding size: {model.config.embedding_size}")
print(f"Model embedding dimension: {model.get_input_embeddings().embedding_dim}")

input_text = "This is a dummy input text."
input_ids = tokenizer.encode(input_text, return_tensors="pt")
labels = input_ids.clone()

input_ids = input_ids.to(model.device)
labels = labels.to(model.device)

with torch.no_grad():
    outputs = model(input_ids, labels=labels)

print("\nModel run completed without errors.")
```

Resulting in:

```
Before resizing:
Vocabulary size: 50280
Embedding size: 50304
Model embedding dimension: 2048

After resizing:
Vocabulary size: 50280
Embedding size: 50282
Model embedding dimension: 2048

Model run completed without errors.
```

Note that this does *not* update the `vocab_size`. The intended purpose of having separate `vocab_size` and `embedding_size` is not entirely clear to me yet, but it seems like the intent is to allow separate `vocab_size` and `embedding_size`, as long as `embedding_size > vocab_size`. A few further options for addressing this issue more completely include:
- update the new method to check whether `config.embedding_size` is defined, and update `vocab_size` if it isn't. This corresponds to how the `wte` layer is defined in the first place ([link](https://github.com/allenai/OLMo/blob/0b835a8d74c4cfb925f47341032f0423a84879a1/olmo/model.py#L1032))
- add a separate function for resizing the vocabulary
- treat vocab size vs. embedding size discrepancies as a separate issue and review the current change as-is

Please let me know if you have any questions or suggestions regarding this PR.


djliden:
@2015aroras @AkshitaB I'm not able to assign reviewers, so flagging for review as requested in #491 . Thanks!

hf_olmo/modeling_olmo.py
```diff
@@ -151,6 +151,44 @@ def tie_weights(self):
         if self.config.weight_tying:
             self.model.transformer.ff_out = self.model.transformer.wte
 
+    def resize_token_embeddings(
+        self, new_num_tokens: Optional[int] = None, pad_to_multiple_of: Optional[int] = None
+    ) -> torch.nn.Embedding:
+        """
```
2015aroras, line 157: This comment seems to be the exact same as the one from the parent class. If it is, then it's not needed here (code editors and IDEs should be able to show the parent class's comment when one hovers over `resize_token_embeddings`). Having the comment somewhat implies that we are defining what the method does, but it is the parent doing it.

hf_olmo/modeling_olmo.py
```diff
@@ -151,6 +151,44 @@ def tie_weights(self):
...
+
+        Return:
+            `torch.nn.Embedding`: Pointer to the input tokens Embeddings Module of the model.
+        """
+        model_embeds = self._resize_token_embeddings(new_num_tokens, pad_to_multiple_of)
+        if new_num_tokens is None and pad_to_multiple_of is None:
+            return model_embeds
+
+        # Update base model and current model config
+        self.config.embedding_size = model_embeds.weight.shape[0]
```
2015aroras, line 184: Could you add a warning for when the new embedding size is less than the vocab size? Something like `"Resizing token embeddings to size <size>, which is less than the vocab size <vocab size> of tokenizer.`

2015aroras (commented):
Vocab size captures how many different tokens a tokenizer can produce. Embedding size captures how many different integers the embedding layer can produce an embedding for. Theoretically speaking, the vocab and embedding sizes would be the same, since the embedding layer should have exactly 1 embedding per token and no extra embeddings.

In practice, not having an embedding size that is a multiple of 128 can can slight performance degradation (broadly speaking, having factors of 2 seems to be beneficial). Thus we use a bigger embedding size than needed, resulting in there being embeddings for some integers that are not valid tokens. Hopefully that makes sense.

The PR itself looks good! Just requesting some small changes.

2015aroras:
> @2015aroras I added a warning if the new embedding size is < the vocab size. I could see an argument for making it an error for consistency with [this](https://github.com/djliden/OLMo/blob/36140c40e7e62dfdf4c2b87ec634b62a6bf8170f/olmo/model.py#L1009), or omitting entirely since it will result in an error as soon as you try to do anything with it anyway. Though I think it also works as-is.

I leaned towards warning instead of error since the user is actively modifying the embedding size. I don't feel strongly about this and resizing embeddings is a relatively uncommon scenario, so any of the 3 options is ok.

> I left the comment alone as it does change vocab size to embedding size (so not the same as base class), but please let me know if you'd like it amended regardless. Thanks! 

I didn't notice the vocab size to embedding size change! I think it's ok to leave it as is, though you could describe how the method differs from `PreTrainedModel.resize_token_embeddings`.

2015aroras (commented):
Just small things, otherwise looking good!

djliden:
@2015aroras Changed the warning format and fixed the indentation. Thanks!

2015aroras (approved):
LGTM! Just fix the code style using Ruff and then you can merge this.

djliden:
@2015aroras done!

djliden:
@2015aroras Thanks for the reviews/feedback/etc! I don't have write access to merge this PR myself—rom my side, both this PR and #503 are ready to go. Please let me know if there's anything else needed, otherwise feel free to merge when you have a chance. Appreciate your help!

PR #511, opened 2024-03-20
Train olmo medium

dirkgr:
I forked this from `train-olmo-large` to make an equivalent 7B config. This will bring it back in.

`train-olmo-large` is starting to feel like the "OLMo 1.5" branch.


epwalsh (approved):
Assuming data paths are exactly the same, LGTM

PR #512, opened 2024-03-21
Fixed assertion message to match actual logic

liaoleo:
The previous assertion message incorrectly indicated that the number of exploded_src must be less than or equal to paths_per_worker, contradicting the actual assertion logic (which ensures the length of exploded_src is at least equal to paths_per_worker). This commit corrects that message to accurately reflect the true intent of the assertion.


2015aroras (approved):
Thank you!

PR #521, opened 2024-03-24
Fix k_norm dimension

davidbrandfonbrener:
The k_norm should be the dimension of the key vector, which is head_dim * effective_n_kv_heads, and not d_model / effective_n_kv_heads.

Also fixing unnecessary assert.


dirkgr (approved):
Thanks! This fell through the cracks when we implemented GQA.

PR #523, opened 2024-03-24
Fix mean reduction in cross entropy loss

davidbrandfonbrener:
The mean reduction should reduce the s_loss to a scalar. Also, I'm not sure why division was being used here instead of multiplication by the mask, but I changed it to multiplication.


epwalsh (approved):
Good catch! Thank you @davidbrandfonbrener

PR #527, opened 2024-03-26
Make the training data properly public

dirkgr:
It is not fast like this, because we make a lot of small range requests per batch, but it works.


epwalsh (approved):
LGTM, also just curious why HTTP and not HTTPS.

PR #532, opened 2024-04-01
Adds a new experimental sharded checkpointer from OLMo-core

epwalsh:
This works a lot like the "torch_new" checkpointer where each rank only downloads the exact amount of data it needs and you can load a checkpoint with a different distributed topology, but without the additional GPU memory overhead.

For now we'll have to rely on installing the OLMo-core package directly from GitHub.


dirkgr (approved):
By itself, this seems good. The main question is, is olmo-core ready for prime-time?

epwalsh:
> The main question is, is olmo-core ready for prime-time?

Probably not, but this will be a good stress test.

PR #533, opened 2024-04-01
make garbage collection interval configurable

epwalsh:


olmo/config.py
```diff
@@ -987,6 +987,11 @@ class TrainConfig(BaseConfig):
     How often to log to the console.
     """
 
+    gen1_gc_interval: int = 1
```
2015aroras, line 990: nit: it might be nice to provide an easy way to turn this off (e.g. setting it to None)

2015aroras (approved):

olmo/config.py
epwalsh: Done! 989f7992

(I assumed you mean disabling manual collection, i.e. not messing with it at all)

dirkgr (approved):

olmo/config.py
```diff
@@ -987,9 +987,10 @@ class TrainConfig(BaseConfig):
     How often to log to the console.
     """
 
-    gen1_gc_interval: int = 1
+    gen1_gc_interval: Optional[int] = 1
```
2015aroras, line 990: Just checking, we want the default behavior to be messing with GC?

2015aroras (approved):

olmo/config.py
epwalsh: Yes I think so. There might be a better default interval for smaller models, but this was a huge improvement for the 70B.

PR #534, opened 2024-04-02
Fix HF pipeline test failure

2015aroras:
Upgrading to transformers 4.39.1 caused a HF pipeline test to start failing (e.g. https://github.com/allenai/OLMo/actions/runs/8444118035/job/23129075811). The issue is a HF bug (https://github.com/huggingface/transformers/issues/29426) and this is a workaround to get around the problem.


dirkgr (approved):
Surprised this doesn't fail some check for an unused parameter?

2015aroras:
> Surprised this doesn't fail some check for an unused parameter?

It causes a linting warning. It can be valid when implementing an inherited method without needing all the parameters. I assume this is why it doesn't cause an error.

PR #537, opened 2024-04-09
Tokenizer patch

AkshitaB:
If you just want to fix the tokenizer files, without running the full conversion:

```python
fix_bad_tokenizer(path)  # assumes native olmo compatibility, i.e., ensure that the original config.yaml is present
write_tokenizer(path)
```

### DIFF
--- a/hf_olmo/convert_olmo_to_hf.py
+++ b/hf_olmo/convert_olmo_to_hf.py
@@ -91,12 +91,14 @@ def download_remote_checkpoint_and_convert_to_hf(checkpoint_dir: str, local_dir:
     convert_checkpoint(local_model_path)
     return local_model_path
 
+
 def fix_bad_tokenizer(checkpoint_dir: str):
     path = os.path.join(checkpoint_dir, "config.yaml")
     conf = om.load(path)
     conf["tokenizer"]["identifier"] = "allenai/gpt-neox-olmo-dolma-v1_5"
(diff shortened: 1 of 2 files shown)


soldni (approved):

OyvindTafjord (approved):
Thanks, LGTM!  At some point, for future-proofing new tokenizer, maybe make some check that the 50279 token is what we expect it to be.

hf_olmo/convert_olmo_to_hf.py
```diff
@@ -91,6 +92,14 @@ def download_remote_checkpoint_and_convert_to_hf(checkpoint_dir: str, local_dir:
     return local_model_path
 
 
+def fix_bad_tokenizer(checkpoint_dir: str):
+    path = os.path.join(checkpoint_dir, "config.yaml")
+    conf = om.load(path)
+    conf["tokenizer"]["identifier"] = "allenai/gpt-neox-olmo-dolma-v1_5"
```
2015aroras, line 98: Just checking, was the tokenizer also wrong for the OLMo models we released, or just for the 1.7 runs? I just want to make sure that this script doesn't break old checkpoints, or that we warn that there is such a risk.

AkshitaB: I ran this explicitly for the v1 models before releasing.

PR #540, opened 2024-04-09
Hybrid sharding with specific number of replicas

2015aroras:
Adds the ability to specify the number of model instances when using a hybrid sharding strategy. If the number of instances is less than the number of nodes, multiple nodes will be used per instance.


epwalsh:
@2015aroras can we get this merged? Then I'll follow up with another PR for the optimizer stuff

epwalsh (approved):

PR #542, opened 2024-04-11
Add cfg option `--scheduler.warmup_min_lr`

epwalsh:
For setting the starting LR during the warmup period. When no set this defaults to the current behavior of starting at 10% of the target LR.

### DIFF
--- a/olmo/optim.py
+++ b/olmo/optim.py
@@ -560,6 +560,7 @@ class BoltOnWarmupScheduler(Scheduler):
             inner=scheduler,
             warmup_start=warmup_start,
             warmup_end=warmup_end,
+            warmup_min_lr=None,
         )
 
     def get_lr(self, initial_lr: float, step: int, max_steps: int) -> float:
--- a/olmo/optim.py
+++ b/olmo/optim.py
@@ -449,6 +449,7 @@ class Scheduler(metaclass=ABCMeta):
     # about how the scheduler subclasses are defined.
     grad_clip_warmup_steps: Optional[int]
     grad_clip_warmup_factor: Optional[float]
+    warmup_min_lr: Optional[float]
 
     @abstractmethod
     def get_lr(self, initial_lr: float, step: int, max_steps: int) -> float:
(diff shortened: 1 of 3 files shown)


dwadden:
Thanks for doing this!

dwadden (approved):
lgtm

olmo/optim.py
```diff
@@ -480,7 +481,11 @@ def get_max_grad_norm_ratio(
         return self._get_max_grad_norm_coeff(initial_max_grad_norm_ratio, step, max_steps)
 
     def _linear_warmup(self, initial_lr: float, step: int, warmup_steps: int = 2000) -> float:
-        return initial_lr * (0.1 + 0.9 * min(step, warmup_steps) / warmup_steps)
+        if self.warmup_min_lr is not None:
+            assert initial_lr > self.warmup_min_lr
```
2015aroras, line 485: nit: You could also add a `assert self.warmup_min_lr >= 0` for good measure

olmo/optim.py
```diff
@@ -480,7 +481,11 @@ def get_max_grad_norm_ratio(
...
+            return self.warmup_min_lr + (initial_lr - self.warmup_min_lr) * min(step, warmup_steps) / warmup_steps
+        else:
```
2015aroras, line 487: From a cleaner code perspective, maybe it would be better to have just 1 formula and use `self.warmup_min_lr` to set the starting lr. Something like:
```
warmup_min_lr = self.warmup_min_lr if self.warmup_min_lr is not None else 0.1 * initial_lr
...
return warmup_min_lr + (initial_lr - warmup_min_lr) * min(step, warmup_steps) / warmup_steps
```

2015aroras (approved):
LGTM! Just minor comments

olmo/optim.py
epwalsh: Done: 75eba56a

olmo/optim.py
epwalsh: Done: 75eba56a

PR #547, opened 2024-04-17
Add OLMo April + July 2024 Checkpoints & Config

2015aroras:


configs/official/OLMo-1.7-7B.yaml
```diff
@@ -0,0 +1,1276 @@
...
+  num_workers: 32
+  drop_last: true
+  pin_memory: true
+  prefetch_factor: 8
+  persistent_workers: true
+  timeout: 0
+  paths:
+    ######### NON WEB DATA #########
+    # ~> GUTENBERG BOOKS (5.256 GT)
+    - s3://ai2-llm/preprocessed/olmo-mix/v1_6-decontaminated/books/gpt-neox-olmo-dolma-v1_5/part-0-00000.npy
```
2015aroras, line 222: `https` data paths not yet available

dumitrac:
@2015aroras - do you think that for 1.7 we should have two configs? I understand that training was stopped, config changed, then resumed.

soheeyang:
Hi, thank you so much for the work! Is there an estimated date of when all the details including the https links of the training data and the data order files are going to be released?

2015aroras:
@kyleclo @soldni Do you have any updates on the data side?

configs/official/OLMo-1.7-7B.yaml
2015aroras: Done

PR #556, opened 2024-04-24
Make hf_olmo support new transformers versions

2015aroras:
`hf_olmo` and the OLMo integration into `transformers` are not compatible, since both try to register a model from the string "olmo". In particular, `hf_olmo` fails to import. This change makes `hf_olmo` not register the model & config when the new transformers version is present. Thus:

- On old versions, `AutoModelForCausalLM.from_pretrained("allenai/OLMo-1B")` will work properly as before (if `hf_olmo` is imported). `AutoModelForCausalLM.from_pretrained("allenai/OLMo-1B-hf")` will break.
- On new versions, `AutoModelForCausalLM.from_pretrained("allenai/OLMo-1B")` will break (none of the weights will get loaded). `AutoModelForCausalLM.from_pretrained("allenai/OLMo-1B-hf")` will work properly. Using `OLMoForCausalLM.from_pretrained("allenai/OLMo-1B")` with `OLMoForCausalLM` from `hf_olmo` will work just like in old versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed `tie_weights` method to a no-op as weight tying is handled in olmo/model.py
 - Fixed the size calculation for qk layer norm
 - Fixed pipeline test failure that occurs due to a bug in transformers version 4.39.1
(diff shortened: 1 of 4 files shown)


epwalsh (approved):
LGTM

PR #562, opened 2024-04-26
Add back original legacy unsharding implementation

2015aroras:
The shared memory implementation for legacy checkpoint unsharding does not seem to be working for everyone. Adding the original implementation back in as the default, with the sharded mem implementation usable via an option to `unshard.py`.


2015aroras:
@dwadden @OyvindTafjord FYI

PR #567, opened 2024-05-03
Reverse weight decay

AkshitaB:
Goal: Perform reverse weight decay on embeddings

Multiply weight_decay factor for the embeddings layer by `(1 - norm(embeddings))`

TODO:

- [ ] What to do when the log metric interval is > 1?

I tried this on a tiny test model config and got an overflow error. Possibly this will not be an issue with the actual model.


olmo/train.py
```diff
@@ -711,6 +711,12 @@ def train_step(self, batch: Dict[str, Any], reduce_global_loss: bool = True) ->
             process_group=self.fsdp_model.process_group,
         )
 
+        # TODO: what to do otherwise?
+        if should_log_optim_metrics_this_step:
+            emb_decay_factor = 1.0 - optim_metrics["param/transformer.wte.weight.norm"]
+        else:
+            emb_decay_factor = 1.0
```
dirkgr, line 718: Norm is not the same as stddev. `sqrt(norm^2 / (embedding_size * vocab size))` is the stddev.

We produce this value every step, because we need it for grad norm. We just don't output it every step. Can you change it so we output it every step, and then we don't have a special case here?

AkshitaB: We compute the norm of the gradient every step (`grad/transformer.wte.weight.norm`), not the norm of the parameter itself (`param/transformer.wte.weight.norm`). Don't we need the latter?

dirkgr:
You are right. Then we need to make sure we compute this every time.

AkshitaB:
> You are right. Then we need to make sure we compute this every time.

Done

olmo/config.py
```diff
@@ -482,6 +482,12 @@ class OptimizerConfig(BaseConfig):
     If not set, defaults to the wandb `log_interval`.
     """
 
+    reverse_embedding_decay: bool = False
+    """
+    Applying weight decay to embeddings may make them too small, potentially causing spikes.
+    Setting this parameter to true is a way of applying "reverse weight decay" to embeddings.
```
dirkgr, line 488: Applies a regularizer to the embeddings that tries to pull them towards a standard deviation of 1.

dirkgr:
@epwalsh , can you look at this as well? This gets all up in your code.

olmo/optim.py
```diff
@@ -43,6 +43,7 @@ def clip_grads_and_collect_metrics(
         global_step: int,
         collect_param_metrics: bool = True,
         process_group: Optional[dist.ProcessGroup] = None,
+        regularize_embeddings: bool = False,
```
dirkgr, line 46: Why is this a parameter to this function? Shouldn't it be just captured in the parameter groups? That's how all the other regularization works.

dirkgr (changes requested):

olmo/train.py
```diff
@@ -737,6 +742,9 @@ def train_step(self, batch: Dict[str, Any], reduce_global_loss: bool = True) ->
                 self.cfg.max_grad_norm_ratio, self.scheduler_current, self.scheduler_max
             )
 
+            if group["name"] == "embedding_decay_group":
+                group["weight_decay"] *= emb_decay_factor
```
dirkgr, line 746: Does't this multiply up `emb_decay_factor` across batches? It feels like this should just be set, not multiplied? Or is there some other bit that resets `group["weight_decay"]` every time?

olmo/train.py
```diff
@@ -720,8 +719,14 @@ def train_step(self, batch: Dict[str, Any], reduce_global_loss: bool = True) ->
             # passing this process group here ensures metrics are reduced correctly when we're using
             # HYBRID sharding.
             process_group=self.fsdp_model.process_group,
+            regularize_embeddings=self.cfg.optimizer.regularize_embeddings,
         )
 
+        emb_norm = optim_metrics["param/transformer.wte.weight.norm"]
+        emb_size = self.cfg.model.embedding_size or self.cfg.model.vocab_size
+        emb_std = math.sqrt(math.pow(emb_norm, 2) / float(emb_size * self.cfg.model.vocab_size))
+        emb_decay_factor = 1.0 - emb_std
```
dirkgr, line 728: If we're using this to plug into the value for WD, that means it needs to be negative when we want to pull _up_ the values. So then it would be `emb_std - 1`?

dirkgr (changes requested):

olmo/train.py
```diff
@@ -720,8 +719,14 @@ def train_step(self, batch: Dict[str, Any], reduce_global_loss: bool = True) ->
             # passing this process group here ensures metrics are reduced correctly when we're using
             # HYBRID sharding.
             process_group=self.fsdp_model.process_group,
+            regularize_embeddings=self.cfg.optimizer.regularize_embeddings,
         )
 
+        emb_norm = optim_metrics["param/transformer.wte.weight.norm"]
+        emb_size = self.cfg.model.embedding_size or self.cfg.model.vocab_size
+        emb_std = math.sqrt(math.pow(emb_norm, 2) / float(emb_size * self.cfg.model.vocab_size))
```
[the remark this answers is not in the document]
dirkgr: Actually, I think this is a big problem. Embeddings will want to be small, so this will push them up. Unused, or rarely used embeddings will never get updated, so they will get bigger and bigger, skewing the calculation of the stddev more and more.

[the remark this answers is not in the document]
AkshitaB: Figuring out which embeddings to exclude from the stddev computation is going to be tricky in the distributed setting though.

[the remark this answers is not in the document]
epwalsh: Thinking out loud here... what if force the unused params to be zero from the beginning? They would still bias standard deviation by as much as they are different from the mean, but they would always be zero.. I think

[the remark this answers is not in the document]
epwalsh: > Can we explicitly make the unused embeddings zero when we load the model?

I think that's our best bet. I can't think of any issues that would introduce in the middle of training. I suspect those parameters are close to 0 anyway due to weight decay and zero gradients.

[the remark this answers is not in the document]
dirkgr: Rare tokens would still be an issue, but not any more than they always are.

PR #569, opened 2024-05-05
Increment torch version

Muennighoff:
See https://github.com/allenai/OLMo-core/pull/14


epwalsh (approved):

Muennighoff:
Do you know why the test is getting cancelled? Seems like I cannot merge it otherwise

epwalsh:
@Muennighoff huh, not sure. The test runner might be running out of memory. I'll see if I can find a work-around.

epwalsh:
@Muennighoff I split off the HF integration tests, which is what's failing now with the same issue. Not sure what's going on there but feel free to merge anyway. We'll figure that out later.

PR #573, opened 2024-05-13
Optionally load trainer state

Muennighoff:
I may be missing some nuances with the checkpointing but can we do sth akin to this PR to avoid trying to load the trainer state when the file is not present?

### DIFF
--- a/olmo/checkpoint.py
+++ b/olmo/checkpoint.py
@@ -516,6 +516,7 @@ class Checkpointer(metaclass=ABCMeta):
         *,
         local_cache: Optional[PathOrStr] = None,
         load_optimizer_state: bool = True,
+        load_trainer_state: bool = True,
     ) -> Dict[str, Any]:
         """
         Restores a checkpoint to the model and optimizer. Returns the remaining trainer state.
@@ -672,6 +673,7 @@ class FullCheckpointer(Checkpointer):
         *,
         local_cache: Optional[PathOrStr] = None,
         load_optimizer_state: bool = True,
+        load_trainer_state: bool = True,
     ) -> Dict[str, Any]:
         with FSDP.state_dict_type(
             fsdp_model,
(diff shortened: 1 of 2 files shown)


olmo/checkpoint.py
```diff
@@ -516,6 +516,7 @@ def restore_checkpoint(
         *,
         local_cache: Optional[PathOrStr] = None,
         load_optimizer_state: bool = True,
+        load_trainer_state: bool = True,
     ) -> Dict[str, Any]:
```
epwalsh, line 520: Looks like the intended return type is now `Optional[Dict[str, Any]]`?
```suggestion
    ) -> Optional[Dict[str, Any]]:
```

epwalsh (commented):
Looks good overall, just need to update the type hints for the return value.

olmo/checkpoint.py
```diff
@@ -672,6 +673,7 @@ def restore_checkpoint(
...
```
Muennighoff, line 677:
```suggestion
        load_trainer_state: bool = True,
    ) -> Optional[Dict[str, Any]]:
```

olmo/checkpoint.py
```diff
@@ -866,6 +870,7 @@ def restore_checkpoint(
...
```
Muennighoff, line 874:
```suggestion
        load_trainer_state: bool = True,
    ) -> Optional[Dict[str, Any]]:
```

olmo/checkpoint.py
```diff
@@ -1556,6 +1564,7 @@ def restore_checkpoint(
...
```
Muennighoff, line 1568:
```suggestion
        load_trainer_state: bool = True,
    ) -> Optional[Dict[str, Any]]:
```

olmo/checkpoint.py
```diff
@@ -1862,6 +1873,7 @@ def restore_checkpoint(
...
```
Muennighoff, line 1877:
```suggestion
        load_trainer_state: bool = True,
    ) -> Optional[Dict[str, Any]]:
```

Muennighoff:
Type checks are still failing --- do you understand why?

epwalsh:
> Type checks are still failing --- do you understand why?

Looks like you'll need to `assert trainer_state is not None` in the `Trainer` when `load_trainer_state=True`.

PR #575, opened 2024-05-13
Add Weka storage support

2015aroras:
Adds support for Weka storage, which is the primary storage in the Jupiter cluster. It might eventually be added a file system, but for now it is exposed as an S3 endpoint. The main changes are:

- Add Weka for `_get_s3_client`, where it is basically treated identically to R2.
- Add a Weka scheme client for `cached_path`. It didn't seem right to put this in `cached_path` since Weka is not a common/universal storage system.


epwalsh (approved):
LGTM

PR #577, opened 2024-05-13
[wip] comparing vanilla torch model and clipping with OLMo FSDP, no_shard and OLMo clipping

ananyahjha93:
image

There is an order of magnitude difference between the losses between the two setups. @dirkgr @epwalsh can you sanity check the OLMo grad_clipping code for FSDP no_shard/DDP?

On 3 batches being sent to the model again and again, the model should overfit and the loss should go to 0. We can see that in the screenshot for vanilla pytorch run.

Comparing the two runs:
1. OLMo model used in both
2. no FSDP vs FSDP-no_shard (32 bit)
3. torch optim and clipping vs OLMo optim and clipping
4. OLMo scheduler used in both

### DIFF
--- a/olmo/optim.py
+++ b/olmo/optim.py
@@ -7,7 +7,7 @@ from typing import Any, Dict, List, Optional, Tuple
 import torch
 import torch.distributed as dist
 import torch.nn as nn
-from torch.distributed.fsdp import FullyShardedDataParallel
+from torch.distributed.fsdp import FullyShardedDataParallel, ShardingStrategy
 from torch.optim.optimizer import Optimizer as OptimizerBase
 
 from . import LayerNormBase
@@ -76,13 +76,12 @@ class Optimizer(OptimizerBase):
         if process_group is not None:
             dst_rank = dist.get_global_rank(process_group, 0)
 
+        #######################################################################
+        # part 1: collect metrics locally
+        #######################################################################
+
         # Collect metrics locally.
         for group in self.param_groups:
-            if is_distributed():
-                # TODO (epwalsh): handle non-sharded params. We don't have any right now but we would
-                # with ReLoRa, for example.
-                assert group.get("sharded", True) is True
-
             for name, p in zip(group["param_names"], group["params"]):
                 name = self._clean_param_name(name)
                 # Always need to collect the norm of gradients for clipping, even if we're not collecting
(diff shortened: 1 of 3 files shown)


dirkgr:
Can you make this a draft PR? I don't think we'll merge this?

dirkgr:
* no FSDP vs FSDP-no_shard (32 bit)
 * torch optim and clipping vs OLMo optim and clipping
 
 Find out which one of these makes the difference?

tests/grad_norm_test.py
```diff
@@ -0,0 +1,254 @@
...
+    # build fsdp config
+    fsdp_model = FSDP(
+        model,
+        sharding_strategy=ShardingStrategy.NO_SHARD,
+        mixed_precision=MixedPrecision(
+            param_dtype=torch.float32,
+            reduce_dtype=torch.float32,
+            buffer_dtype=torch.float32,
+        ),
+        auto_wrap_policy=None,
```
dirkgr, line 217: Did you check how it wraps the model?

tests/grad_norm_test.py
```diff
@@ -0,0 +1,254 @@
...
+
+    del model
+    del torch_optimizer
+    del scheduler
+    del data_loader
+
+    # use same model, data, optimizer, fsdp_model and send to trainer and compare gradient clip
+
+    # olmo optimizer
+    model = OLMo(cfg.model).to('cuda')
```
dirkgr, line 194: Can you make double sure the initialization is same here?

tests/grad_norm_test.py
```diff
@@ -0,0 +1,254 @@
...
+    olmo_optimizer = build_optimizer(cfg, model)
+    data_loader = build_train_dataloader(cfg)
```
dirkgr, line 196: Make sure the data loader is same too? By the time you get here you might have consumed some random state, so you might get different data. I would go so far as to pre-load the data into a `List`, and then use that, so you can be 100% sure it's the same.

tests/grad_norm_test.py
epwalsh: From the looks of it, this model will definitely start with different weights than the one above. The best way to be sure would be to load starting state dict of the first model into the second model.

olmo/optim.py
```diff
@@ -141,10 +140,17 @@ def clip_grads_and_collect_metrics(
...
+        #######################################################################
+        param_group_sharded = False
+        for group in self.param_groups:
+            param_group_sharded = param_group_sharded or group.get("sharded", False)
+
         # Now reduce metrics over all ranks.
         total_grad_norm: torch.Tensor
         per_param_avg_metrics: List[torch.Tensor] = []
-        if is_distributed():  # TODO (epwalsh): skip for non-sharded params
+        if is_distributed() and param_group_sharded:
```
epwalsh, line 153: Will still need to reduce gradient metrics with non-sharded params since each rank will have different gradients.

ananyahjha93: won't the gradients sync after loss.backward() call?

epwalsh: Ah, yes you're right. My mistake

ananyahjha93: in DDP loss.backward() syncs grads, optimizer.step() updates each copy with synced grads! Or am I getting this wrong? So, if loss.backward() syncs grads, then every rank must have the same grads

ananyahjha93:
image

Sending torch optimizer and gradient clipping with no_shard seems to fix it!

olmo/optim.py
ananyahjha93: @epwalsh does this apply to FSDP no_shard as well?

PR #580, opened 2024-05-16
Update ignore_index parameter for flash attention

2015aroras:
The flash attention library changed the `cross_entropy_loss` parameter `ignored_index` to `ignore_index` in v2.5.8. This changes lets us support v2.5.8.

### DIFF
--- a/olmo/train.py
+++ b/olmo/train.py
@@ -151,12 +151,13 @@ class Trainer:
                 cross_entropy_loss,
             )
 
+            # The `ignored_index` parameter of `cross_entropy_loss` was changed to `ignore_index` in v2.5.8 with commit https://github.com/Dao-AILab/flash-attention/commit/ec6d22143b5d375e253b2ebfc563b26a43f43684
+            ce_loss_use_ignore_index_param = version.parse(flash_attn.__version__) >= version.parse("2.5.8")
+
             def fused_loss_fn(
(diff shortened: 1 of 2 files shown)


olmo/train.py
```diff
@@ -145,22 +146,30 @@ class Trainer:
...
             def fused_loss_fn(
                 logits, labels, ignore_index: int = -100, reduction: str = "mean", compute_z_loss: bool = False
             ):
+                ignore_index_kwarg = {}
+                if version.parse(flash_attn.__version__) >= version.parse("2.5.8"):
+                    # The parameter name was changed in v2.5.8 with commit https://github.com/Dao-AILab/flash-attention/commit/ec6d22143b5d375e253b2ebfc563b26a43f43684
+                    ignore_index_kwarg = {"ignore_index": ignore_index}
+                else:
+                    ignore_index_kwarg = {"ignored_index": ignore_index}
+
```
epwalsh, line 163: I think it would be more efficient to parse the version outside of this function, above, so we don't do it on every call.

epwalsh (commented):
That's annoying! Just one suggestion

olmo/train.py
2015aroras: [74d3efe](https://github.com/allenai/OLMo/pull/580/commits/74d3efe8be29b61f42411ddbf0ccce3dd76a0265)

epwalsh (approved):

PR #582, opened 2024-05-17
Allow hybrid sharding to have multiple replicas in a node

2015aroras:
This PR changes our hybrid sharding to allow multiple replicas in a node. One benefit of this is that we can use hybrid sharding to do no sharding by having the number of replicas equal with the world size. Then we don't need to use DDP or `NO_SHARD` (which is getting deprecated).


epwalsh (approved):
LGTM

PR #593, opened 2024-05-23
Inspect training data without data indices

2015aroras:
This PR updates the `inspect_train_data.py` script to enable inspecting training data when the device data indices are not present. Our runs save these indices locally but not in remote storage. The implementation has the following advantages:

- We can look at training data without the data indices (the main point of this PR).
- The script can be called with an S3 path when data indices are not present (nothing needs to be downloaded locally beforehand).
- The script can look as far into the run as you would like without a perf hit.

The implementation is such that the script will default to the original behavior when data indices are present.

### DIFF
--- a/scripts/inspect_train_data.py
+++ b/scripts/inspect_train_data.py
@@ -47,6 +47,10 @@ def inspect_data_without_device_data_indices(
     )
     cfg.data.num_workers = 1
 
+    if cfg.global_train_batch_size % world_size != 0:
+        raise ValueError(f"World size must divide global_train_batch_size {cfg.global_train_batch_size}")
+    cfg.device_train_batch_size = cfg.global_train_batch_size // world_size
+
     try:
         trainer_state = load_state_dict(run_path, f"step{reference_step}/train/rank0.pt", map_location="cpu")
     except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


scripts/inspect_train_data.py
```diff
@@ -2,28 +2,121 @@
...
+        os.environ["RANK"] = "0"
+        dataloader = build_train_dataloader(cfg, world_size=world_size)
+
+        for rank in ranks:
+            os.environ["RANK"] = str(rank)
+            # Set FS_LOCAL_RANK to a non-zero number so that global data indices are not rewritten
+            os.environ["FS_LOCAL_RANK"] = "1"
+
+            for step in steps:
+                dataloader = build_train_dataloader(cfg, world_size=world_size)
```
epwalsh, line 79: This would have to rebuild the indices file every time, right? That could be slow, but we could probably avoid rebuilding for every rank.

2015aroras: Setting `FS_LOCAL_RANK=1` avoids rebuilding the indices file every time since it's only down for local FS rank 0.
`# Set FS_LOCAL_RANK to a non-zero number so that global data indices are not rewritten`

epwalsh (approved):
LGTM

scripts/inspect_train_data.py
2015aroras: [d137a1f](https://github.com/allenai/OLMo/pull/593/commits/d137a1fe93785487382564883603d98736bf6d15)

PR #601, opened 2024-05-30
Fixes clipping

ananyahjha93:
Added tests (CPU and GPU) to compare torch clipping and olmo clipping and fixed clipping for DDP and FSDP no_shard

@epwalsh can we merge this PR so that I can push the DDP one after this?

### DIFF
--- a/olmo/optim.py
+++ b/olmo/optim.py
@@ -234,7 +234,7 @@ class Optimizer(OptimizerBase):
             clipping_rate = torch.tensor(0.0, device="cpu")
         all_metrics["clipping_rate"] = clipping_rate
 
-        # per_param_norm, clipping_rate and total_grad_norm are computed even without collect_param_metrics set to False
+        # per_param_norm, clipping_rate and total_grad_norm are computed even with collect_param_metrics set to False
         # return those values
         return all_metrics
 
--- a/olmo/optim.py
+++ b/olmo/optim.py
@@ -7,7 +7,7 @@ from typing import Any, Dict, List, Optional, Tuple
 import torch
 import torch.distributed as dist
 import torch.nn as nn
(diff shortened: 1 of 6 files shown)


olmo/optim.py
```diff
@@ -195,27 +199,31 @@ def is_grad_norm_metric(metric_name: str) -> bool:
...
+        #######################################################################
+        # part 3: clip grads
+        #######################################################################
         num_grads_clipped = 0
         num_eligible_grads = 0
         for group in self.param_groups:
             if (max_norm_ratio := group.get("max_grad_norm_ratio")) is not None:
                 num_clipped = self._do_adaptive_clipping(
-                    group, max_norm_ratio, global_step, all_metrics, collect_param_metrics=collect_param_metrics
+                    group, max_norm_ratio, global_step, all_metrics, collect_param_metrics=True
```
epwalsh, line 222: Why this change? This will force a host device sync on every batch.

olmo/optim.py
```diff
@@ -195,27 +199,31 @@ def is_grad_norm_metric(metric_name: str) -> bool:
...
                 )
             elif (max_norm := group.get("max_grad_norm")) is not None:
                 num_clipped = self._do_global_fixed_clipping(
-                    group, max_norm, all_metrics, collect_param_metrics=collect_param_metrics
+                    group, max_norm, all_metrics, collect_param_metrics=True
```
epwalsh, line 226: Same question here.

epwalsh (commented):
@ananyahjha93 please don't merge anything into `main` without a review. This PR touches some critical components and I'm pretty sure your changes will force an unnecessary host device sync on every batch, which could potentially have a big negative impact on training throughput.

olmo/optim.py
ananyahjha93: let me send in a PR for this fix!

PR #603, opened 2024-05-31
Unshard without passing checkpointer type

2015aroras:
The unsharding script currently requires us to pass in the checkpoint type, but this is not necessary since it can be read from config. Passing in the type has been a nuisance in the past, so this PR changes the script to default to the checkpoint type in the config. It changes the unsharder to use `build_sharded_checkpointer`, so that we don't need to duplicate that logic in the unsharding script.

Currently untested.


epwalsh (approved):
Nice

2015aroras:
This worked fine for a checkpoint created with the `local` checkpointer.

PR #604, opened 2024-06-03
Adds a tool that diffs two wandb runs

dirkgr:


AkshitaB (changes requested):
Minor suggestion: it will be great if it takes in https paths that we directly copy from the browser, and processes the strings to remove them to pass to the api.run command. A bit quicker for the user.

scripts/compare_wandb_configs.py
```diff
@@ -0,0 +1,69 @@
...
+
+import click
+
+from olmo.util import prepare_cli_environment
+
+log = logging.getLogger(__name__)
+
+
+def flatten(dictionary, parent_key="", separator="."):
+    items = []
```
2015aroras, line 12: Just curious, why not use a dictionary and `dict.update(...)` instead of a list with `extend()`?

scripts/compare_wandb_configs.py
```diff
@@ -0,0 +1,69 @@
...
+        print("Settings only in right:")
+        print("\n".join(f"\t{k}: {right_config[k]}" for k in sorted(right_only_keys)))
+        print()
+
+    keys_with_differences = {
+        k for k in left_config.keys() & right_config.keys() if left_config[k] != right_config[k]
+    }
+    if len(keys_with_differences) > 0:
+        if len(left_only_keys) > 0 or len(right_only_keys) > 0:
+            print("Settings with differences:")
```
2015aroras, line 61: Having this in an if statement is a bit odd, but I guess you prefer it that way?

2015aroras (approved):
Approved with nits

scripts/compare_wandb_configs.py
dirkgr: It prints this heading if it already printed other headings. Otherwise, there is only one heading, which we don't really need.

scripts/compare_wandb_configs.py
dirkgr: Because I copy and pasted this from SO 🙈.

Fixed in 049969b22d0b6d0789ae2fc48f2b40111f9ff533.

2015aroras (approved):

dirkgr:
@AkshitaB , I added your suggestion in 598a885.

AkshitaB (approved):

PR #606, opened 2024-06-04
now accepts wandb project and entity as options

drschwenk:
I seemed to need to supply project and entity when initializing the wandb Api. Otherwise defaulted to another wandb entity.


dirkgr:
A run id looks like this: `ai2-llm/olmo-medium/765zdauu`. It already contains project and entity. Why do you need to specify them extra?

drschwenk:
Ah, slight terminology thing- I was thinking run_id (just the last unique id portion), not the full run_path.

PR #607, opened 2024-06-10
Rewrite initialization

AkshitaB:
Simplifies our inscrutable initialization

* IMPORTANT: currently, the implementation matches the old buggy values for init in several places. See below.
* Removes `init_weights` with its complex if-else logic.
* Adds `init_normal` which only takes the module, the std, and optionally a cutoff_factor.
* std and cutoff_factor computation is now handled in each module's `reset_parameters()`
* Adds unit tests for initialization.
* Removes implementation for `kaiming_normal` and `fan_in` InitFnType as these aren't being used anywhere. Can be added later if needed.

Potential bugs found in initialization as a result of the refactoring (these will be fixed after feedback):

- [x] `OLMoBlock.ff_out`'s `normal` initialization [multiples std by an extra factor](https://github.com/allenai/OLMo/pull/607/files#diff-ef8ab7279deeec716e70a1cc9ab2accaaa60f27b301cc0733f1e00a9e39c07d1R483-R490) of `1 / math.sqrt(2 * self.config.n_layers`. This potentially came from trying to incorporate `full_megatron` into the same function. 
- [x] Hardcoded values: `mitchell` hardcodes a cutoff_factor of 3.0 (always truncated_normal_ with 3.0). `full_megatron` hardcodes a default cutoff_factor of 3.0 (truncated_normal_ with `config.init_cutoff_factor or 3.0`). Again, this may be a result of trying to incorporate multiple inits into the same function. Ideally, the cutoff_factor should always come from the configurable `config.init_cutoff_factor`; do we _want_ to set always this value to 3.0 for mitchell and megatron?
- [ ] Need clarification: Why do we [scale the embedding](https://github.com/allenai/OLMo/pull/607/files#diff-ef8ab7279deeec716e70a1cc9ab2accaaa60f27b301cc0733f1e00a9e39c07d1R1071) with the following factor if `scale_logits=True`?
        `emb_std_factor = (0.5 * math.sqrt(self.config.d_model)) if self.config.scale_logits else 1.0`
- [ ] Additionally, in case of `mitchell` init, due to supplying the factor at multiple places in the old code, [std ends up always being 0.5](https://github.com/allenai/OLMo/pull/607/files#diff-ef8ab7279deeec716e70a1cc9ab2accaaa60f27b301cc0733f1e00a9e39c07d1R1077) when `scale_logits=True`!

### DIFF
--- a/olmo/model.py
+++ b/olmo/model.py
@@ -480,13 +480,11 @@ class OLMoBlock(nn.Module):
             cutoff_factor = self.config.init_cutoff_factor
 
         elif self.config.init_fn == InitFnType.mitchell:
-            # TODO: 3 is currently hardcoded; this should be using init_cutoff_factor instead.
             attn_out_std = 1 / (math.sqrt(2 * self.config.d_model * (self.layer_id + 1)))
             ff_out_std = 1 / (math.sqrt(2 * self.ff_out.in_features * (self.layer_id + 1)))
-            cutoff_factor = 3.0
+            cutoff_factor = self.config.init_cutoff_factor or 3.0
 
         elif self.config.init_fn == InitFnType.full_megatron:
-            # TODO: 3 is hardcoded as default value
             attn_out_std = ff_out_std = self.config.init_std / math.sqrt(2.0 * self.config.n_layers)
             cutoff_factor = self.config.init_cutoff_factor or 3.0
 
@@ -676,7 +674,7 @@ class OLMoSequentialBlock(OLMoBlock):
             cutoff_factor = self.config.init_cutoff_factor
         elif self.config.init_fn == InitFnType.mitchell:
             std = 1 / math.sqrt(self.config.d_model)
-            cutoff_factor = 3.0
+            cutoff_factor = self.config.init_cutoff_factor or 3.0
         elif self.config.init_fn == InitFnType.full_megatron:
             std = self.config.init_std
             cutoff_factor = self.config.init_cutoff_factor or 3.0
@@ -791,7 +789,7 @@ class OLMoLlamaBlock(OLMoBlock):
             cutoff_factor = self.config.init_cutoff_factor
         elif self.config.init_fn == InitFnType.mitchell:
             std = 1 / math.sqrt(self.config.d_model)
-            cutoff_factor = 3.0
+            cutoff_factor = self.config.init_cutoff_factor or 3.0
         elif self.config.init_fn == InitFnType.full_megatron:
             std = self.config.init_std
             cutoff_factor = self.config.init_cutoff_factor or 3.0
@@ -1059,7 +1057,7 @@ class OLMo(nn.Module):
         elif self.config.init_fn == InitFnType.mitchell:
             # TODO: this is buggy! std will always be 0.5 when scale_logits = True
             wte_std = emb_std_factor / math.sqrt(self.config.d_model)
-            wte_cutoff_factor = 3.0
+            wte_cutoff_factor = self.config.init_cutoff_factor or 3.0
         elif self.config.init_fn == InitFnType.full_megatron:
             wte_std = self.config.init_std
             wte_cutoff_factor = self.config.init_cutoff_factor or 3.0
@@ -1074,7 +1072,7 @@ class OLMo(nn.Module):
                 wpe_cutoff_factor = self.config.init_cutoff_factor
             elif self.config.init_fn == InitFnType.mitchell:
                 wpe_std = 1 / math.sqrt(self.config.d_model)
-                wpe_cutoff_factor = 3.0
+                wpe_cutoff_factor = self.config.init_cutoff_factor or 3.0
             elif self.config.init_fn == InitFnType.full_megatron:
                 wpe_std = self.config.init_std
                 wpe_cutoff_factor = self.config.init_cutoff_factor or 3.0
@@ -1093,7 +1091,7 @@ class OLMo(nn.Module):
                 ff_out_cutoff_factor = self.config.init_cutoff_factor
             elif self.config.init_fn == InitFnType.mitchell:
                 ff_out_std = 1 / math.sqrt(self.config.d_model)
-                ff_out_cutoff_factor = 3.0
+                ff_out_cutoff_factor = self.config.init_cutoff_factor or 3.0
             elif self.config.init_fn == InitFnType.full_megatron:
                 ff_out_std = 1 / math.sqrt(self.config.d_model)
                 ff_out_cutoff_factor = self.config.init_cutoff_factor or 3.0
--- a/olmo/model.py
+++ b/olmo/model.py
@@ -961,12 +961,6 @@ class OLMo(nn.Module):
         if self.config.alibi and self.config.rope:
             raise OLMoConfigurationError("ALiBi and RoPE are mutually exclusive")
 
-        if self.config.scale_logits:
-            if self.config.weight_tying:
-                raise OLMoConfigurationError("`scale_logits` and `weight_tying` are mutually exclusive")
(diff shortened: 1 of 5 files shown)


olmo/model.py
```diff
@@ -473,20 +474,42 @@ def reset_parameters(self):
...
+            init_normal(self.ff_out, std=self.config.init_std, init_cutoff_factor=self.config.init_cutoff_factor)
+
+            # TODO: what is up with this extra divisor? We may not want this.
+            # TODO: This potentially came from trying to include full_megatron init.
+            # git blame rabbit hole:
+            #   https://github.com/allenai/OLMo/pull/239/commits/0c5b7b5ccac26c3e4729322520f1fdd5b0a4d54e#diff-14cdafe7342088f0c4851cc670185044902088ba6bc5ddcf0a464d698e4165fe
+            #   https://github.com/allenai/OLMo/commit/ba20a857fde0c2fb2e8dbcba07e3781fe088ba06
+            # This is for the case of ff_out._is_residual = True
+            with torch.no_grad():
+                self.ff_out.weight.div_(math.sqrt(2 * self.config.n_layers))
```
epwalsh, line 489: This was copied from the original MPT code, and now I'm thinking we should remove this unless there's precedent somewhere else, like megatron.

olmo/model.py
```diff
@@ -473,20 +474,42 @@ def reset_parameters(self):
...
-            type_of_module=ModuleType.out_module,
-        )
-        init_weights(
-            self.config,
-            self.ff_out,
-            d=self.ff_out.in_features,
-            layer_id=self.layer_id,
-            type_of_module=ModuleType.out_module,
-        )
+
```
epwalsh, line 477: nit: we call `init_normal` in every block here on the same modules, so I think it would be cleaner to structure this in the same way as you've structured it in `OLMoSequentialBlock`, where you just set the `std` and `cutoff_factor` in the `if-elif` blocks.

olmo/model.py
```diff
@@ -1020,21 +1065,57 @@ def device(self) -> torch.device:
...
         if hasattr(self.transformer, "ff_out"):
-            init_weights(self.config, self.transformer.ff_out, type_of_module=ModuleType.final_out)  # type: ignore
+            if self.config.init_fn == InitFnType.normal:
+                ff_out_std = self.config.init_std
+                ff_out_cutoff_factor = self.config.init_cutoff_factor
+            elif self.config.init_fn == InitFnType.mitchell:
+                ff_out_std = 1 / math.sqrt(self.config.d_model)
+                ff_out_cutoff_factor = 3.0
+            elif self.config.init_fn == InitFnType.full_megatron:
+                ff_out_std = self.config.d_model**-0.5
```
epwalsh, line 1113: For consistency:
```suggestion
                ff_out_std = 1 / math.sqrt(self.config.d_model)
```

olmo/model.py
```diff
@@ -1020,21 +1065,57 @@ def device(self) -> torch.device:
...
+            else:
+                raise NotImplementedError(self.config.init_fn)
+
+            init_normal(self.transformer.wpe, std=wpe_std, init_cutoff_factor=wpe_cutoff_factor)
 
         # Top-level layer norm.
         self.transformer.ln_f.reset_parameters()  # type: ignore
 
         # Output weights.
         if hasattr(self.transformer, "ff_out"):
```
epwalsh, line 1105: It seems odd that with untied weights we don't change the std dev of `ff_out` when `scale_logits=True`. This might be another bug.

olmo/model.py
```diff
@@ -473,20 +474,42 @@ def reset_parameters(self):
...
+            # TODO: This potentially came from trying to include full_megatron init.
+            # git blame rabbit hole:
+            #   https://github.com/allenai/OLMo/pull/239/commits/0c5b7b5ccac26c3e4729322520f1fdd5b0a4d54e#diff-14cdafe7342088f0c4851cc670185044902088ba6bc5ddcf0a464d698e4165fe
+            #   https://github.com/allenai/OLMo/commit/ba20a857fde0c2fb2e8dbcba07e3781fe088ba06
+            # This is for the case of ff_out._is_residual = True
+            with torch.no_grad():
+                self.ff_out.weight.div_(math.sqrt(2 * self.config.n_layers))
+
+        elif self.config.init_fn == InitFnType.mitchell:
+            # TODO: 3 is currently hardcoded; this should be using init_cutoff_factor instead.
```
epwalsh, line 492: How about just default to 3 when it's `None` like we do with megatron init?

epwalsh (changes requested):
No major concerns. I'm glad we're cleaning this up.

> Why do we [scale the embedding](https://github.com/allenai/OLMo/pull/607/files#diff-ef8ab7279deeec716e70a1cc9ab2accaaa60f27b301cc0733f1e00a9e39c07d1R1071) with the following factor if scale_logits=True?
emb_std_factor = (0.5 * math.sqrt(self.config.d_model)) if self.config.scale_logits else 1.0.

This was another "trick" we heard works from someone else (not sure who).

AkshitaB:
> No major concerns. I'm glad we're cleaning this up.
> 
> > Why do we [scale the embedding](https://github.com/allenai/OLMo/pull/607/files#diff-ef8ab7279deeec716e70a1cc9ab2accaaa60f27b301cc0733f1e00a9e39c07d1R1071) with the following factor if scale_logits=True?
> > emb_std_factor = (0.5 * math.sqrt(self.config.d_model)) if self.config.scale_logits else 1.0.
> 
> This was another "trick" we heard works from someone else (not sure who).

Wouldn't this make more sense if we did this when `weight_tying` was on? I'm trying to get a sense of intuition for some of these choices/tricks.

olmo/model.py
AkshitaB: Yes, I planned to do it after we cleared up the confusion regarding the extra multiplier.

epwalsh:
> > No major concerns. I'm glad we're cleaning this up.
> > > Why do we [scale the embedding](https://github.com/allenai/OLMo/pull/607/files#diff-ef8ab7279deeec716e70a1cc9ab2accaaa60f27b301cc0733f1e00a9e39c07d1R1071) with the following factor if scale_logits=True?
> > > emb_std_factor = (0.5 * math.sqrt(self.config.d_model)) if self.config.scale_logits else 1.0.
> > 
> > 
> > This was another "trick" we heard works from someone else (not sure who).
> 
> Wouldn't this make more sense if we did this when `weight_tying` was on? I'm trying to get a sense of intuition for some of these choices/tricks.

Yea I'm guessing that's the only scenario where we tried it? It might have come from PaLM.

olmo/model.py
AkshitaB: Where does `3.0` come from?

epwalsh: That's probably what Mitchell had hardcoded.

epwalsh (approved):
LGTM

olmo/model.py
```diff
@@ -933,6 +961,12 @@ def __init__(self, config: ModelConfig, init_params: bool = True):
         if self.config.alibi and self.config.rope:
             raise OLMoConfigurationError("ALiBi and RoPE are mutually exclusive")
 
+        if self.config.scale_logits:
+            if self.config.weight_tying:
+                raise OLMoConfigurationError("`scale_logits` and `weight_tying` are mutually exclusive")
+            if self.config.init_fn != InitFnType.normal:
+                raise OLMoConfigurationError("`scale_logits` can only be used with `normal` init.")
```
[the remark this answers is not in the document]
epwalsh: Disregard above -- discussed offline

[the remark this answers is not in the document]
AkshitaB: Done, fixed

PR #612, opened 2024-06-11
adding DDP to the codebase

ananyahjha93:


ananyahjha93:
Need to add a test to check DDP works well with checkpointing.

epwalsh (changes requested):
Can you revert the unrelated formatting changes and remove the empty configs?

ananyahjha93:
@epwalsh done!

ananyahjha93:
at the 300M size, it follows the FSDP model loss curve, being better initially, but both lines showing the same trend after sometime

image

olmo/train.py
```diff
@@ -657,36 +685,35 @@ def train_batch(self, batch: Dict[str, Any]) -> Tuple[torch.Tensor, Optional[tor
...
+            # setup sync context for DDP for all micro-batches except the last
+            ddp_context = nullcontext
+            if (
+                self.cfg.distributed_strategy == DistributedStrategy.ddp
+                and self.cfg.ddp.grad_sync_mode == DDPGradSyncMode.batch
+            ):
+                if micro_batch_idx != num_micro_batches - 1:
+                    ddp_context = self.dist_model.no_sync
+
+            with ddp_context():
```
epwalsh, line 700: Nit: `ddp_context` is a confusing name for this variable. I would suggest `grad_sync_context`:

```suggestion
            grad_sync_context = nullcontext
            if (
                self.cfg.distributed_strategy == DistributedStrategy.ddp
                and self.cfg.ddp.grad_sync_mode == DDPGradSyncMode.batch
            ):
                if micro_batch_idx != num_micro_batches - 1:
                    grad_sync_context = self.dist_model.no_sync

            with grad_sync_context():
```

olmo/train.py
```diff
@@ -763,9 +790,13 @@ def train_step(self, batch: Dict[str, Any], reduce_global_loss: bool = True) ->
             metrics["train/ZLoss"] = z_batch_loss.item()
 
         # Maybe collect post-step optimizer-specific metrics.
+        # TODO (ananya): self.dist_model with DDP will not work with ``get_post_step_metrics`` of LionW
+        # while AdamW does not implement a custom version of ``get_post_step_metrics`` function. Since we do not use
+        # LionW anymore, we can skip fixing the interaction of DDP model and ``get_post_step_metrics`` function
+        # in LionW for the time being.
```
epwalsh, line 796: Instead of putting this comment here, I would suggest throwing an `assert isinstance(model, FSDP)` in `LionW.get_post_step_metrics()`. That way this fails loudly.

olmo/config.py
```diff
@@ -49,6 +49,8 @@
     "WandbConfig",
     "CompilerConfig",
     "WandbConfig",
+    "DistributedStrategy",
+    "DDPGradSyncMode",
```
epwalsh, line 53: Missing `DDPConfig` I believe:
```suggestion
    "DDPGradSyncMode",
    "DDPConfig",
```

olmo/config.py
```diff
@@ -1038,11 +1090,21 @@ class TrainConfig(BaseConfig):
...
+
     fsdp: FSDPConfig = field(default_factory=FSDPConfig)
     """
     Fully sharded data parallel settings.
     """
 
+    ddp: DDPConfig = field(default_factory=DDPConfig)
+    """
+    DDP settings.
+    """
```
epwalsh, line 1106: I would make these optional now, otherwise we're going to look at some configs later and be like "why are there DDP settings in this config? I thought we ran this FSDP"
```suggestion
    fsdp: Optional[FSDPConfig] = field(default_factory=FSDPConfig)
    """
    Fully sharded data parallel settings.
    """

    ddp: Optional[DDPConfig] = None
    """
    DDP settings.
    """
```

scripts/tiny/gantry-script.sh
epwalsh, line 1: Can you put these scripts in `scripts/beaker/tiny/...` instead?

scripts/train.py
```diff
@@ -124,69 +130,89 @@ def main(cfg: TrainConfig) -> None:
...
 
-    if version.parse(torch.__version__) >= version.parse("2.1.0"):
-        # This prevents any parameters from being initialized twice
-        def dummy_init_fn(module: torch.nn.Module) -> None:
-            module.to_empty(device=get_default_device())
+        if cfg.ddp.find_unused_params is True:
+            assert (
+                cfg.ddp.grad_sync_mode == DDPGradSyncMode.micro_batch
+            ), "`find_unused_params` is set to True. DDP needs to synchronize gradients for every micro-batch to avoid errors. Setting `grad_sync_mode` to `micro_batch`."
+            cfg.ddp.grad_sync_mode = DDPGradSyncMode.micro_batch
```
epwalsh, line 147: With the assert above we'd never get here.

scripts/train.py
```diff
@@ -124,69 +130,89 @@ def main(cfg: TrainConfig) -> None:
...
+        ), "DDP does not work with init_device set to anything other than `cuda`."
 
-    if version.parse(torch.__version__) >= version.parse("2.1.0"):
-        # This prevents any parameters from being initialized twice
-        def dummy_init_fn(module: torch.nn.Module) -> None:
-            module.to_empty(device=get_default_device())
+        if cfg.ddp.find_unused_params is True:
+            assert (
+                cfg.ddp.grad_sync_mode == DDPGradSyncMode.micro_batch
+            ), "`find_unused_params` is set to True. DDP needs to synchronize gradients for every micro-batch to avoid errors. Setting `grad_sync_mode` to `micro_batch`."
```
epwalsh, line 146: Can you throw `OLMoConfigurationError` instead of `AssertionError`? That's our convention.

scripts/train.py
```diff
@@ -212,9 +238,30 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
...
+            if cfg.distributed_strategy == DistributedStrategy.ddp:
+                checkpoint_type = CheckpointType.unsharded
+
+                try:
+                    assert cfg.save_interval_unsharded is not None
+                except AssertionError:
+                    log.info(
+                        "DDP requires setting `save_interval_unsharded`. Using the value set for `save_interval`."
+                    )
+                    cfg.save_interval_unsharded = cfg.save_interval
```
epwalsh, line 250: This could just be an `if` block instead of using error handling.

epwalsh: I would also log this at WARNING level instead of INFO.

scripts/train.py
```diff
@@ -212,9 +238,30 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
...
+
+                try:
+                    assert cfg.save_num_unsharded_checkpoints_to_keep > 0
+                except AssertionError:
+                    log.info(
+                        "DDP requires setting `save_num_unsharded_checkpoints_to_keep`. Using the value set for `save_num_checkpoints_to_keep`."
+                    )
+                    cfg.save_num_unsharded_checkpoints_to_keep = cfg.save_num_checkpoints_to_keep
```
epwalsh, line 258: Same here.

scripts/train.py
```diff
@@ -212,9 +238,30 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
...
+            elif cfg.distributed_strategy == DistributedStrategy.fsdp:
+                checkpoint_type = (
+                    CheckpointType.sharded if cfg.save_num_checkpoints_to_keep != 0 else CheckpointType.unsharded
+                )
+            else:
+                log.info(f"Distributed strategy {cfg.distributed_strategy} not supported yet!")
```
epwalsh, line 264: If it's not supported we should be throwing an error, no?

olmo/checkpoint.py
```diff
@@ -672,92 +669,144 @@ def save_checkpoint(
...
-            except FileNotFoundError:
-                # for backwards compatibility
-                trainer_state = load_state_dict(load_path, "other.pt", local_cache=local_cache)
+            gc.collect()
+            torch.cuda.empty_cache()
+            barrier()
+        else:
+            log.info(
+                "`FullCheckpointer.restore_checkpoint` only supported for FSDP and DDP distributed strategies!"
+            )
```
epwalsh, line 769: This should throw an error

epwalsh (changes requested):
This adds a lot of complexity to a trainer that's already become a little too complex. Can you remind me why you need DDP over just FSDP with NO_SHARD?

olmo/checkpoint.py
```diff
@@ -611,49 +609,48 @@ class FullCheckpointer(Checkpointer):
...
+
+                    # Then the optimizer state.
+                    optim_state_dict = FSDP.optim_state_dict(dist_model, optim)
+                    self._write_optim_dict(
+                        optim_state_dict, checkpoint_dir, upload_to, save_overwrite=self.cfg.save_overwrite
                     )
-                del optim_state_dict
-                barrier()
+            elif isinstance(dist_model, DDP):
+                # First, get the model state dict from DDP wrapped model
```
dirkgr, line 639: Check whether we need to do something to prevent all the ranks from writing this file?

olmo/checkpoint.py
```diff
@@ -672,92 +669,144 @@ def save_checkpoint(
...
+                    barrier()
+                    for turn in range(get_local_world_size()):
+                        log.info("Loading optimizer state turn %d ...", turn)
+                        if turn == get_local_rank():
+                            load_fsdp_optim_state(dist_model, optim, optim_state_dict_to_load)
+                            gc.collect()
+                            torch.cuda.empty_cache()
+                        barrier()
+                    del optim_state_dict_to_load
+        elif isinstance(dist_model, DDP):
```
dirkgr, line 748: Check whether there is an easy way to send a model from rank0 to the other ranks.

scripts/train.py
ananyahjha93: should I throw a configuration error or quietly correct this?

```
if cfg.ddp.find_unused_params is True and cfg.ddp.grad_sync_mode != DDPGradSyncMode.micro_batch:
            cfg.ddp.grad_sync_mode = DDPGradSyncMode.micro_batch
            log.warning("`find_unused_params` is set to True. DDP needs to synchronize gradients for every micro-batch to avoid errors. Setting `grad_sync_mode` to `micro_batch`.") 
```

ananyahjha93: ah we should probably raise an error so that we maintain uniformity and information in the config files.

olmo/checkpoint.py
ananyahjha93: _write_model_dict and _write_optim_dict have a check for this!

```
if get_global_rank() == 0:
```

olmo/checkpoint.py
ananyahjha93: cant find an easy fix, maybe @epwalsh knows something here!

scripts/train.py
```diff
@@ -124,69 +130,91 @@ def main(cfg: TrainConfig) -> None:
...
-        def dummy_init_fn(module: torch.nn.Module) -> None:
-            module.to_empty(device=get_default_device())
+    if cfg.distributed_strategy == DistributedStrategy.ddp:
+        log.info("Wrapping model with DDP...")
+        assert cfg.ddp is not None, "DistributedStrategy ddp needs cfg.ddp to be set!"
 
-        param_init_fn = dummy_init_fn
-    else:
-        param_init_fn = None
+        # raise error instead of quite correcting, this will help maintain information across configs
```
epwalsh, line 141: No need to comment when the error message itself is pretty clear IMO
```suggestion
```

olmo/checkpoint.py
epwalsh: No easy way I'm aware of.

olmo/checkpoint.py
```diff
@@ -1836,10 +1912,13 @@ def save_checkpoint(
             save_model_and_optim_state,
         )
 
+        assert isinstance(
+            dist_model, FSDP
+        ), f"{self.__class__.__name__} is being called to save a model where `distributed_strategy` is not FSDP."
```
epwalsh, line 1917: Actually olmo-core can handle this fine

olmo/checkpoint.py
```diff
@@ -1872,8 +1951,11 @@ def restore_checkpoint(
             load_model_and_optim_state,
         )
 
+        assert isinstance(
+            dist_model, FSDP
+        ), f"{self.__class__.__name__} is being called to load a model where `distributed_strategy` is not FSDP."
```
epwalsh, line 1956:
```suggestion
```

olmo/checkpoint.py
epwalsh:
```suggestion
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added clipping fix to `Optimizer` class to make it work with FSDP `no_shard` and DDP.
 - Added tests to compare grad norm differences between torch optimizer and clipping and OLMo optimizer and clipping on both CPU and GPU.
 - Expose memmap dtype in data config 
+- Added DDP to the codebase.
```
epwalsh, line 15:
```suggestion
- Added support for DDP training.
```

epwalsh (changes requested):
LGTM, just a few minor comments left.

epwalsh (approved):

PR #615, opened 2024-06-11
Officially add OLMo-core as a dependency

epwalsh:


pyproject.toml
```diff
@@ -15,6 +15,7 @@ license = { file = "LICENSE" }
 dependencies = [
     "numpy",
     "torch>=2.1,<2.3",
+    "ai2-olmo-core==0.1.0",
```
2015aroras, line 18: Since we are maintaining OLMo-core, setting a version lower bound could be an option.
```suggestion
    "ai2-olmo-core>=0.1.0",
```

2015aroras (approved):
It's nice to have this at last!

pyproject.toml
epwalsh: I'd like to hold off on that until it becomes somewhat stable. Not necessarily v1.0 stable, but I think it's still too soon.

PR #616, opened 2024-06-11
Chameleon stability experiments

AkshitaB:
Main changes:

Norm reordering (`norm_after` config option).
The zloss multiplier is now configurable (`auxiliary_loss_multiplier` config option).


olmo/train.py
```diff
@@ -131,7 +136,59 @@ def cross_entropy_loss(
...
+def fused_loss_fn(
+    logits,
+    labels,
+    ignore_index: int = -100,
+    reduction: str = "mean",
+    compute_z_loss: bool = False,
+    z_loss_multiplier: float = 1e-4,
+):
+    import flash_attn
+    from flash_attn.ops.triton.cross_entropy import cross_entropy_loss  # type: ignore
```
epwalsh, line 153: I worry about importing modules within a function that's called repetitively. Alternatively you could throw this whole definition within a `try import flash_attn: ... except ImportError` block. That would probably be more efficient at runtime.

olmo/model.py
```diff
@@ -1058,6 +1082,8 @@ def reset_parameters(self):
             wte_cutoff_factor = self.config.init_cutoff_factor or 3.0
         elif self.config.init_fn == InitFnType.full_megatron:
             wte_std = self.config.init_std
+            if self.config.scale_emb_init:
+                wte_std *= math.sqrt(self.config.d_model)
```
epwalsh, line 1086: Should we apply this to other init methods as well? The docstring for this option makes it seem like it would be applied to all of them.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added config options for `norm_after` and `auxiliary_loss_multiplier` (used with zloss).
```
epwalsh, line 12:
```suggestion
- Added config options for `model.norm_after`, `model.scale_emb_init`, and `auxiliary_loss_multiplier` (used with zloss).
```

epwalsh (changes requested):
Looks good overall, only minor comments

olmo/model.py
AkshitaB: We had discussed this a bit at the time of https://github.com/allenai/OLMo/pull/607 , ultimately deciding that it was not something that was important to decide right away. I've added a clarification in the docstring.

tests/fused_loss_test.py
epwalsh, line 1: Can you rename this to `train_test.py` since it's testing code in `train.py`?

tests/fused_loss_test.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+import torch
+from torch.testing import assert_close
+
+from olmo.train import cross_entropy_loss, fused_loss_fn
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Requires CUDA device")
```
epwalsh, line 8: You'll also need to mark this as a GPU test, otherwise it doesn't run in CI
```suggestion
@pytest.mark.skipif(torch.cuda.device_count() < 1, reason="Requires CUDA device")
@pytest.mark.gpu
```

epwalsh (approved):

PR #620, opened 2024-06-12
Running Amber experiments at 7B

epwalsh:


configs/llm-360-amber-s3.yaml
```diff
@@ -0,0 +1,508 @@
+run_name: llm-360-amber-data-hyper-init-repro-3
```
drschwenk, line 1: Just to be consistent, do we want to call this amberish7, or amberish7-base, something like that (I think this name predates the amberish moniker)?

configs/llm-360-amber-s3.yaml
```diff
@@ -0,0 +1,508 @@
...
+scheduler:
+  name: cosine_with_warmup
+  units: tokens
+  t_warmup: 9175040000
+  t_max: 1.25e12
+  alpha_f: 0.1
+  warmup_min_lr: 0.0
+
+tokenizer:
+  identifier: tokenizers/nousresearch-llama-2-7b-hf.json
```
drschwenk, line 66: I've since changed this to match ambers: 

huggyllama/llama-7b

This was the tokenizer I used for my pile experiment. While it's close, it's not identical to ambers.

configs/llm-360-amber-s3.yaml
```diff
@@ -0,0 +1,508 @@
...
+sharded_checkpointer: olmo_core
+
+save_interval_unsharded: null
+save_num_unsharded_checkpoints_to_keep: -1
+
+load_path: null
+
+max_duration: 2ep
+global_train_batch_size: 2240
+device_train_microbatch_size: 2
```
drschwenk, line 85: I didn't change this originally, but should we try to up this since our sequence length is now half the size?

configs/llm-360-amber-s3.yaml
```diff
@@ -0,0 +1,508 @@
...
+  #       pile-validation:
+  #       - s3://ai2-llm/eval-data/perplexity/v3_small_gptneox20b/pile/val/part-0-00000.npy
+  #       wikitext_103-validation:
+  #         - s3://ai2-llm/eval-data/perplexity/v3_small_gptneox20b/wikitext_103/val/part-0-00000.npy
+
+  ##########################
+  # Downstream evaluations #
+  ##########################
+
+  - label: hellaswag
```
drschwenk, line 135: I put in the full downstream evals in the 1B.

scripts/beaker/epwalsh-stability/llm-360-amber.sh
```diff
@@ -0,0 +1,59 @@
...
+NUM_NODES=$1
+shift
+
+BEAKER_REPLICA_RANK=$1
+shift
+
+# Setup Python environment.
+conda shell.bash activate base
+
+# Install flash-attn
```
drschwenk, line 18: I'm not sure if this is from your stability experiments. I haven't been doing this... not sure if I should be?

tokenizers/nousresearch-llama-2-7b-hf.json
drschwenk, line 1: We can remove this when changing the tokenizer.

scripts/beaker/epwalsh-stability/llm-360-amber-launch.sh
```diff
@@ -0,0 +1,41 @@
...
+  --host-networking \
+  --budget ai2/oe-training \
+  --no-nfs \
+  --propagate-failure \
+  --synchronized-start-timeout 30m \
+  --no-python \
+  --env LOG_FILTER_TYPE=local_rank0_only \
+  --env OMP_NUM_THREADS=8 \
+  --env OLMO_TASK=model \
+  --env-secret AWS_ACCESS_KEY_ID=DUSTINS_AWS_ACCESS_KEY_ID \
```
drschwenk, line 27: I know it doesn't really matter, but I've been changing these for my runs (just so things don't suddenly break if someone deletes their key from the shared workspace).

drschwenk (commented):
Just had a few comments/questions.

scripts/beaker/epwalsh-stability/llm-360-amber.sh
epwalsh: You might already have flash in your image.

configs/llm-360-amber-s3.yaml
epwalsh: done

configs/llm-360-amber-s3.yaml
epwalsh: Might be able to get away with that but my per-device batch isn't divisible by 4, I want to keep global batch size the same for now.

PR #623, opened 2024-06-13
HF dataset loading optimizations

2015aroras:
Issue: Loading HF datasets for downstream evals has been slowing down the start of our runs because:
1. Every process tries to load each HF dataset at the same time. This both results in a lot of network traffic to one endpoint (potentially resulting in throttling?), and also maybe some contention over HF's dataset cache on disk.
2. Even when a dataset has already been cached locally, trying to load the dataset results in network calls as HF checks for changes to the data.

Fix: This PR tackles these issues by:
1. Utilizing HF's `load_from_disk` and `save_to_disk` dataset methods to save a local copy of the datasets. These datasets are no longer associated with the online versions, and so loading them from disk does not result in network traffic (as far as I can tell).
2. Making only the FS rank 0 process perform network calls for HF dataset loading. Doing this requires coordination between the processes using `barrier()`, but this seems to be less problematic. Note that barriers are invoked each time a dataset needs to be loaded (on the positive side, if all datasets are already in the cache then there are no barrier invocations).

In a 2 GPU interactive session, the performance goes from 10 mins without this new cache to 1 min with this cache ready beforehand. I have populated a copy of the cache already at `/net/weka/reviz/hf_datasets_cache`.

I haven't yet verified that the evaluation correctness is not affected by this change (because no GPUs available), but I don't expect this to negatively impact eval correctness.

Edit: Now the barrier will always be invoked exactly once per dataset

### DIFF
--- a/olmo/eval/__init__.py
+++ b/olmo/eval/__init__.py
@@ -32,7 +32,9 @@ def build_downstream_evaluator(
     task_class = label_to_task_map[eval_cfg.label]
     if isinstance(task_class, tuple):
         task_class, task_kwargs = task_class
-    ds_eval_dataset = task_class(tokenizer=tokenizer, **task_kwargs)  # type: ignore
+    ds_eval_dataset = task_class(
+        tokenizer=tokenizer, datasets_cache_dir=train_config.hf_datasets_cache_dir, **task_kwargs
+    )  # type: ignore
     data_config = eval_cfg.data
     if is_unit_test:
         ds_eval_sampler = None
--- a/olmo/eval/__init__.py
+++ b/olmo/eval/__init__.py
@@ -34,7 +34,7 @@ def build_downstream_evaluator(
         task_class, task_kwargs = task_class
     ds_eval_dataset = task_class(
         tokenizer=tokenizer, datasets_cache_dir=train_config.hf_datasets_cache_dir, **task_kwargs
-    )  # type: ignore
+    )
     data_config = eval_cfg.data
     if is_unit_test:
         ds_eval_sampler = None
--- a/olmo/eval/__init__.py
+++ b/olmo/eval/__init__.py
@@ -34,7 +34,7 @@ def build_downstream_evaluator(
         task_class, task_kwargs = task_class
     ds_eval_dataset = task_class(
         tokenizer=tokenizer, datasets_cache_dir=train_config.hf_datasets_cache_dir, **task_kwargs
-    )
+    )  # type: ignore
     data_config = eval_cfg.data
     if is_unit_test:
         ds_eval_sampler = None
--- a/olmo/eval/downstream.py
+++ b/olmo/eval/downstream.py
@@ -9,9 +9,60 @@ import torch.nn.functional as F
 from sklearn.metrics import f1_score
 from torchmetrics import Metric
 
+from olmo.torch_util import barrier, get_fs_local_rank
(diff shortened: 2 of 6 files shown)


olmo/eval/downstream.py
```diff
@@ -9,11 +9,62 @@
...
 
+from olmo.torch_util import barrier, get_fs_local_rank
+from olmo.util import load_dataset_from_disk, save_dataset_to_disk
+
 from ..tokenizer import Tokenizer
 
 log = logging.getLogger(__name__)
 
 
+def load_dataset(path, name, split, datasets_cache_dir: Optional[str] = None):
```
epwalsh, line 20: nit: missing type hints
```suggestion
def load_dataset(path, name, split, datasets_cache_dir: Optional[str] = None):
```

olmo/eval/__init__.py
```diff
@@ -32,7 +32,9 @@ def build_downstream_evaluator(
     task_class = label_to_task_map[eval_cfg.label]
     if isinstance(task_class, tuple):
         task_class, task_kwargs = task_class
-    ds_eval_dataset = task_class(tokenizer=tokenizer, **task_kwargs)  # type: ignore
+    ds_eval_dataset = task_class(
+        tokenizer=tokenizer, datasets_cache_dir=train_config.hf_datasets_cache_dir, **task_kwargs
+    )  # type: ignore
```
epwalsh, line 37: I think this `# type: ignore` is probably on the wrong line now

olmo/eval/downstream.py
```diff
@@ -9,11 +9,62 @@
...
+                "Path %s name %s split %s not present in local dir %s, loading from online",
+                path,
+                name,
+                split,
+                datasets_cache_dir,
+            )
+        # Barrier here to stop the case where FS rank 0 saves the dataset to disk before some non-zero
+        # ranks try getting the dataset from disk. This would cause those non-zero ranks to bypass
+        # the next barrier and cause rank 0 to be stuck at that barrier.
+        barrier()
```
epwalsh, line 36: This seems a little sketch, but I guess it's fine as long as all ranks either have or do not have the dataset on disk, which is probably a reasonable assumption?

epwalsh (approved):
LGTM, only minor comments

olmo/eval/downstream.py
2015aroras: [c55974d](https://github.com/allenai/OLMo/pull/623/commits/c55974d5581d0cf44e86e996c7ec4a86b97d7966)

olmo/eval/downstream.py
2015aroras: [9f2d9d2](https://github.com/allenai/OLMo/pull/623/commits/9f2d9d25aa61f749f545fbd1a058f70227838f8c)
I've changed the logic to be simpler and more robust against this case. Now the idea is that FS local rank 0 does its thing first (load from disk or online and cache), then every other rank follows afterwards. This is less optimized but simpler and safer imo.

olmo/eval/__init__.py
2015aroras: Intellisense puts it there and the type checking passes with it there (and fails without it there).

olmo/eval/downstream.py
epwalsh: I like that much better ✅

PR #624, opened 2024-06-14
Make olmo-core checkpointer more robust on weka

epwalsh:


olmo/checkpoint.py
```diff
@@ -1838,9 +1844,7 @@ def save_checkpoint(
 
         with self._temporary_wd(dir) as checkpoint_dir:
             log.info("Saving model and optim state...")
-            local_files_created = save_model_and_optim_state(
-                checkpoint_dir, fsdp_model, optim, save_overwrite=self.cfg.save_overwrite
-            )
+            local_files_created = save_model_and_optim_state(checkpoint_dir, fsdp_model, optim)
```
2015aroras, line 1847: Why remove `save_overwrite`?

epwalsh: We handle `save_overwrite` in this checkpointer class already. At this point in the code we're passing a temporary directory that we've guaranteed exists and is empty, so if we pass `save_overwrite=True`, then olmo-core will delete and recreate that directory, which can cause issues on weka/nfs.

2015aroras (approved):

PR #626, opened 2024-06-17
Inspect training data improvements

2015aroras:
- Add option to pass in directories mounted in runs, so that they can be undone in the script
- Indicate if an instance is masked


epwalsh (approved):
LGTM

Issue #584, opened 2024-05-18
Problems with multi-epoch training


Muennighoff:
### 🐛 Describe the bug

I think there are two problems with multi-epoch training:

- Training finishes if setting e.g. `duration: 2e12T` & 1 epoch < 2e12 tokens. It currently requires setting `duration: 2ep` but it should also work with `T` I think (also mentioned here: https://github.com/allenai/OLMo/issues/554)
olmo.train:816	INFO	[step=817847/1430511]
train/CrossEntropyLoss=2.341
train/Perplexity=10.39
throughput/total_tokens=1,715,149,471,744
throughput/device/tokens_per_second=18,573
throughput/device/batches_per_second=0.5668
olmo.train:1172	INFO	Training epoch complete
olmo.train:1194	INFO	Saving final checkpoint...
train:238	INFO	Training complete
- Afaict when resuming a run in >1 epoch state, it requires newly setting `epoch: num_epochs` in the config to ensure that the data is in a different order: https://github.com/allenai/OLMo/blob/e6430a070e97154c5b3b5fc35a7416026e9353ad/olmo/data/__init__.py#L103
I think we should just load this from the trainer state dict. However, afaict this is currently not happening because the checkpoint is only loaded after the IterableDataset is already created. I.e. data loader is loaded:
https://github.com/allenai/OLMo/blob/e6430a070e97154c5b3b5fc35a7416026e9353ad/scripts/train.py#L116
Checkpoint with epoch value is loaded:
https://github.com/allenai/OLMo/blob/e6430a070e97154c5b3b5fc35a7416026e9353ad/scripts/train.py#L238
& the data loader remains unchanged.

 Without knowing this, people will train the 2nd epoch with the same data order as the 1st.

### Versions

latest main

PR #628, opened 2024-06-18
Olmo tiny scripts

ananyahjha93:


scripts/train.py
```diff
@@ -248,7 +248,7 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
                     )
                     cfg.save_interval_unsharded = cfg.save_interval
 
-                if cfg.save_num_unsharded_checkpoints_to_keep < 1:
+                if cfg.save_num_unsharded_checkpoints_to_keep == 0:
                     log.warning(
```
AkshitaB, line 252: What if `save_num_checkpoints_to_keep` is also 0?

AkshitaB (commented):
Discussed my queries offline with @ananyahjha93 

* _How were model shapes decided?_ Based on Pythia and then number of parameters.
* _How about LR?_ Also ballpark from Pythia. 

Other things to note:
* Global batch size may also require some ablation

configs/tiny/OLMo-300M.yaml
```diff
@@ -9,17 +9,15 @@ wandb:
 model:
   d_model: 1024
   n_heads: 16
-  n_layers: 16
+  n_layers: 24
   mlp_ratio: 8
   weight_tying: false
   alibi: false
   rope: true
   flash_attention: true  # not available on AMD
```
dirkgr, line 17: Is now available on AMD

configs/tiny/OLMo-300M.yaml
```diff
@@ -133,31 +139,19 @@ evaluators:
...
-
   - label: commonsense_qa
     type: downstream
 
   - label: social_iqa
     type: downstream
 
-  - label: basic_arithmetic
-    type: downstream
-
```
dirkgr, line 160: What's wrong with these?

configs/tiny/OLMo-300M.yaml
```diff
@@ -47,47 +47,53 @@ optimizer:
...
-save_num_unsharded_checkpoints_to_keep: 3
+save_num_unsharded_checkpoints_to_keep: -1
 
 load_path: null
 
-max_duration: 100_000  # 419B tokens, this is for the scheduler
-stop_at: 100_000
-global_train_batch_size: 2048
-device_train_microbatch_size: 8
+max_duration: 2ep
```
dirkgr, line 81: This means you'll run into this bug: https://github.com/allenai/OLMo/issues/584
It might not matter. The problem is only that the second epoch will be shuffled the same way the first one is shuffled.

configs/tiny/OLMo-300M.yaml
```diff
@@ -9,17 +9,15 @@ wandb:
 model:
```
dirkgr, line 9: No DDP section in this file?

configs/tiny/OLMo-20M.yaml
```diff
@@ -0,0 +1,1262 @@
...
+  metrics_log_interval: 10
+
+scheduler:
+  name: cosine_with_warmup
+  units: tokens
+  t_warmup: 4194304000
+  t_max: 3e12
+  alpha_f: 0.1
+  grad_clip_warmup_steps: null
+  grad_clip_warmup_factor: 5
```
dirkgr, line 64: Don't have these settings.

configs/tiny/OLMo-20M.yaml
```diff
@@ -0,0 +1,1262 @@
...
+  num_workers: 32
+  drop_last: true
+  pin_memory: true
+  prefetch_factor: 8
+  persistent_workers: true
+  timeout: 0
+  paths:
+    ######### NON WEB DATA #########
+    # ~> GUTENBERG BOOKS (5.256 GT)
+    - s3://ai2-llm/preprocessed/olmo-mix/v1_6-decontaminated/books/gpt-neox-olmo-dolma-v1_5/part-0-00000.npy
```
dirkgr, line 214: Can you read from weka instead?

configs/tiny/OLMo-300M.yaml
```diff
@@ -47,47 +47,53 @@ optimizer:
...
+remote_save_folder: s3://ai2-llm/checkpoints/olmo-tiny/300M/${run_name}
 save_overwrite: false
-# Sharded checkpoints (best for restarts)
-save_interval: 5000
-save_num_checkpoints_to_keep: 3
+
 # Unsharded checkpoints (for ddp)
 save_interval_unsharded: 5000
-save_num_unsharded_checkpoints_to_keep: 3
+save_num_unsharded_checkpoints_to_keep: -1
```
dirkgr, line 77: What does `-1` do?

dirkgr (changes requested):
I assume the configs between the sizes are all the same, so I didn't look at all of them.

scripts/train.py
ananyahjha93: it then assumes that you did not want to keep checkpoints at all!

configs/tiny/OLMo-300M.yaml
ananyahjha93: ah, ``basic_arithmetic`` should be in, others don't provide any signal based on my experience

configs/tiny/OLMo-300M.yaml
ananyahjha93: removed the comment

configs/tiny/OLMo-300M.yaml
ananyahjha93: there is!

scripts/train.py
ananyahjha93: -1 assumes you want to save all checkpoints and so I made it ``==0`` instead of ``< 1``.

configs/tiny/OLMo-20M.yaml
ananyahjha93: was planning to run on pluto, now I can see free nodes on jupiter, making the change!

configs/tiny/OLMo-300M.yaml
ananyahjha93: I'll add a ``stop_at`` 400k steps!

configs/tiny/OLMo-300M.yaml
ananyahjha93: ah this was commented out saying

``# Doesn't work from cache.``

configs/tiny/OLMo-20M.yaml
ananyahjha93: took these from @AkshitaB 's ``llamaish1-normal-weka.yaml.``

configs/tiny/OLMo-300M.yaml
ananyahjha93: -1 is for keeping all checkpoints, but I'll double check

configs/tiny/OLMo-20M.yaml
ananyahjha93: removed them for now!

configs/tiny/OLMo-300M.yaml
AkshitaB: Should work with cache v4

configs/tiny/OLMo-150M.yaml
```diff
@@ -56,20 +56,16 @@ optimizer:
 
 scheduler:
   name: cosine_with_warmup
-  units: tokens
-  t_warmup: 4194304000
-  t_max: 3e12
+  t_warmup: 5000
```
dirkgr, line 59: For normal init, this is a lot of warmup? Not a big deal, but unusual?

configs/tiny/OLMo-150M.yaml
```diff
@@ -78,9 +74,10 @@ save_num_unsharded_checkpoints_to_keep: -1
 
 load_path: null
 
-max_duration: 2ep
+max_duration: 1ep
+stop_at: 406_934
```
dirkgr, line 78: Do you need both `max_duration` and `stop_at`?

configs/tiny/OLMo-150M.yaml
```diff
@@ -152,6 +149,10 @@ evaluators:
   - label: social_iqa
     type: downstream
 
+  # Doesn't work from cache.
+  # - label: basic_arithmetic
+  #   type: downstream
+
```
dirkgr, line 155: Even with cache v4?

scripts/beaker/tiny/torchrun-script.sh
```diff
@@ -26,9 +26,10 @@ torchrun \
...
   --rdzv_endpoint=$BEAKER_LEADER_REPLICA_HOSTNAME:29400 \
   scripts/train.py \
-    configs/tiny/OLMo-30M.yaml \
+    configs/tiny/OLMo-20M.yaml \
       --run_name=$TASK_NAME \
       --wandb.name=$TASK_NAME \
       --wandb.group=$TASK_NAME \
-      --wandb.project=tiny_olmo \
+      --wandb.project=olmo-tiny \
+      --max_grad_norm=2.0 \
```
dirkgr, line 34: Do you want to do this clipping value for all small models?

dirkgr (approved):
Approved with a small comment about the long warmup.

configs/tiny/OLMo-150M.yaml
ananyahjha93: yes, from what I have observed and Dave mentioned the training goes past max_duration if stop_at is not set

configs/tiny/OLMo-150M.yaml
ananyahjha93: smaller models, higher LR, did not take a chance! never bad doing a longer warmup!

scripts/beaker/tiny/torchrun-script.sh
ananyahjha93: ah, let me fix this, so the model with clipping value 2.0 does not show any downstream improvement!

olmo/train.py
```diff
@@ -82,6 +95,11 @@ def reset(self) -> None:
 
     def check(self) -> Dict[str, float]:
         metrics: Dict[str, float] = {"throughput/total_tokens": self.global_total_tokens}
+
+        # plot flops related metrics
+        metrics["throughput/total_training_Gflops"] = self.total_training_Gflops
+        metrics["throughput/total_training_log_Gflops"] = math.log(self.total_training_Gflops)
```
epwalsh, line 101: Why not just `throughput/Gflops`?
```suggestion
        metrics["throughput/Gflops"] = self.total_training_Gflops
        metrics["throughput/log_Gflops"] = math.log(self.total_training_Gflops)
```

epwalsh: I guess "total" is meant to convey that it's forward plus backward flops, but for other metrics we use "total" to mean "cumulative throughout training," like with `throughput/total_tokens`.

ananyahjha93: yeah, so the idea is that training flops consumed until this step!

epwalsh: Oh, but that would reset on every restart. What do you use this metric for?

ananyahjha93: is there a ``global_batch_seen_until_now`` metric?

ananyahjha93: I guess, I can do ``total_tokens`` / (``seq_len`` * ``global_batch_size``)

ananyahjha93: ah, lets refactor ``num_fwf_flops`` and ``num_bck_flops`` to be per token and then a simple multiplication with ``global_total_tokens`` solves everything!

olmo/train.py
```diff
@@ -1094,10 +1111,12 @@ def on_trace_ready(p):
...
                     self.global_train_tokens_seen += global_batch_size * seq_len
                     speed_monitor.batch_start(
-                        self.global_train_tokens_seen,
-                        batch_size * seq_len,  # num tokens in batch for this device
+                        global_total_tokens=self.global_train_tokens_seen,
+                        device_batch_num_tokens=batch_size * seq_len,  # num tokens in batch for this device
                         # We start monitoring speed after the first batch since the first
                         # batch might be an outlier due to compiling and other initialization overhead.
+                        num_fwd_flops=self.model.num_fwd_flops,  # this is per sequence
+                        num_bck_flops=self.model.num_bck_flops,  # this is per sequence
```
epwalsh, line 1119: "this is per sequence" ... it's per-token now, right?

ananyahjha93: changed

epwalsh (approved):

PR #631, opened 2024-06-21
Make hf_olmo support AutoModelForCausalLM

2015aroras:
`hf_olmo` was conflicting with `transformers` because they were both claiming `model_type=olmo`. Changing `hf_olmo`'s model type to something else seems to make all our problems go away...

We need to go change all our uploaded models' checkpoints before we merge this though, but thankfully having the wrong model type in config before this gets merged is not a breaking issue (only causes warnings).


epwalsh (approved):
👍

PR #636, opened 2024-06-28
Add option to skip optim steps for 0 grad params

epwalsh:
#605 should be reviewed and merged first.

This PR adds the ability to skip optimizer updates for the parts of parameters that have 0 gradients, such as the embeddings for tokens not present in the current batch (assuming no weight tying).


olmo/optim.py
```diff
@@ -419,21 +429,30 @@ def step(self, closure=None) -> None:
...
-                # Perform step weight decay
-                p.data.mul_(1 - group["lr"] * group["weight_decay"])
-
-                grad = p.grad
                 state = self.state[p]
 
+                # Perform step weight decay
+                mask: Optional[torch.Tensor] = None
+                if self._selective_updates:
+                    mask = grad != 0
```
2015aroras, line 451: thought: you could instead do `mask = grad != 0 if self._selective_updates else 1`, and assume the mask is always present in subsequent logic.

olmo/optim.py
```diff
@@ -373,9 +376,12 @@ def __init__(
         super().__init__(params, defaults)
         for group in self.param_groups:
             group["initial_lr"] = group["lr"]
+        self._selective_updates = selective_updates
```
2015aroras, line 379: nit: Like in the other PR, this could be moved into the parent class

2015aroras (approved):

olmo/optim.py
epwalsh: good call: 1024122f

olmo/optim.py
```diff
@@ -510,16 +512,20 @@ def step(self, closure=None) -> None:
 class AdamW(torch.optim.AdamW, Optimizer):
     def __init__(self, *args, record_update_metrics: bool = False, selective_updates: bool = False, **kwargs):
         super().__init__(*args, **kwargs)
-        self._record_step_size = record_update_metrics
+
+        # Need to set these here just like in our base `Optimizer` class since our `Optimizer.__init__`
```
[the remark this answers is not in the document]
epwalsh: yea this gets messy b/c our `Optimizer.__init__()` also calls PyTorch's `Optimizer.__init__()`, which would then get called twice here unless we didn't call `torch.optim.AdamW.__init__()` (via `super().__init__()`), but then we'd have to copy over all the other code that happens within `torch.optim.AdamW.__init__()`.

PR #638, opened 2024-06-28
Scaling laws pipeline

AkshitaB:
Scripts for power law fitting


scripts/scaling/predict_olmo_tiny_examples.py
```diff
@@ -0,0 +1,135 @@
...
+
+# from olmo.scaling.scaling_laws.utils import validation
+
+CONFIGS = {
+    "mup-olmo-128-train": {
+        "path": "wandb_outputs/mup-olmo-128-train.csv",
+        "keys": ["train/CrossEntropyLoss"],  # validation
+        "train_step_min": 150,
+        "train_step_max": 550,
+        "eval_step_max": 750,
```
liujch1998, line 26: These numbers look pretty small. Is that intentional in MUP? My experience is that early loss points can be noisy, possibly due to LR warmup, so I didn't include points prior to 5k or 10k steps. Does MUP have LR warmup?

liujch1998 (approved):

scripts/scaling/predict_olmo_tiny_examples.py
AkshitaB: No, this is just an example to test that the code works.

olmo/scaling/scaling_laws/extrapolate_n.py
```diff
@@ -0,0 +1,70 @@
...
+
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy
+
+from .curve_fit import get_coefficients
+
+
+@dataclass
+class CurveFitConfig:
```
AkshitaB, line 13: @liujch1998 We don't need 3 different versions of the CurveFitConfig, do we? It can all be rolled into one

scripts/scaling/extrapolate_n.py
```diff
@@ -0,0 +1,104 @@
...
+    CurveFitConfig,
+    get_data_at_d,
+    plot_n_scaling_at_d,
+)
+from olmo.scaling.scaling_laws.utils import validation
+
+VAL_KEYS = [f'eval/{val}/CrossEntropyLoss' for val in validation]
+
+CONFIG_BY_N = {
+    21266432: {
```
AkshitaB, line 21: We should have more human-readable names here

olmo/scaling/scaling_laws/extrapolate_n.py
```diff
@@ -0,0 +1,70 @@
...
+                    y = yy
+            if y is None: # there is no data at or later than d tokens
+                continue
+            if config.mode == 'train':
+                train_ns.append(n)
+                train_ys.append(y)
+            elif config.mode == 'eval':
+                eval_ns.append(n)
+                eval_ys.append(y)
+    return train_ns, train_ys, eval_ns, eval_ys
```
AkshitaB, line 58: Isn't this a pared-down version of what's in the `curve_fit.py` file already? Can we combine these two files?

olmo/scaling/scaling_laws/extrapolate_n.py
liujch1998: The config needed for running extrapolation on N is quite different from that of extrapolating D (i.e. curve_fit.py), so I thought it'd be nice to keep them separate so it's clear what is required. I can rename them into distinct class names.

scripts/scaling/extrapolate_n.py
liujch1998: `CONFIG_BY_MODEL_SIZE`?

olmo/scaling/scaling_laws/extrapolate_n.py
liujch1998: Details are a bit different. I feel it'd be more work to consolidate them than keeping them separate.

Plus I think we should rename `curve_fit.py` to `extrapolate_d.py` to distinguish it from extrapolate_n ...

scripts/scaling/extrapolate_n.py
AkshitaB: I mean, the keys. They can be "20M", "1B" etc. ?

liujch1998: Thanks! I have taken care of this.

AkshitaB:
@liujch1998 I'm closing this in favor of `ladder-1xC` branch so that we don't have to work on two separate branches. I've merged all the changes from this branch into that one, so that all the scaling / power laws related code is consolidated.

PR #639, opened 2024-06-30
MoE

Muennighoff:
Replaces https://github.com/allenai/OLMo/pull/541


olmo/model.py
```diff
@@ -660,10 +660,245 @@ def build(cls, layer_id: int, config: ModelConfig, cache: BufferCache) -> OLMoBl
...
+    """
+
+    def __init__(self, layer_id: int, config: ModelConfig, cache: BufferCache):
+        try:
+            from megablocks.layers.arguments import Arguments as MoEArgs
+            from megablocks.layers.dmoe import dMoE
+            from megablocks.layers.moe import MoE
+        except ImportError:
+            raise ImportError(
+                "To train MoEs, run `pip install git+https://github.com/Muennighoff/megablocks.git@olmoe`"
```
epwalsh, line 682: What's different about your branch for the original source?

Muennighoff: Mainly that it includes zloss which we use during training for better stability

Muennighoff: you can view the exact difference here: https://github.com/databricks/megablocks/compare/main...Muennighoff:megablocks:olmoe ; besides zloss it also has expert choice which is currently not used but i think we may want to try in the future when we go multimodal

olmo/config.py
```diff
@@ -457,6 +462,61 @@ class ModelConfig(BaseConfig):
     See :data:`TrainConfig.precision` instead.
     """
 
+    moe_num_experts: Optional[int] = 8
+    """
+    The number of experts to use in the MoE block.
+    """
+
+    moe_top_k: Optional[int] = 2
```
dirkgr, line 470: If these are `Optional`, what does it mean when it's `None`?

olmo/initialization.py
```diff
@@ -13,9 +13,15 @@ def init_normal(
     # weights
     if init_cutoff_factor is not None:
         cutoff_value = init_cutoff_factor * std
-        nn.init.trunc_normal_(module.weight, mean=0.0, std=std, a=-cutoff_value, b=cutoff_value)
+        if hasattr(module, "weight"):
+            nn.init.trunc_normal_(module.weight, mean=0.0, std=std, a=-cutoff_value, b=cutoff_value)
+        else:
+            nn.init.trunc_normal_(module, mean=0.0, std=std, a=-cutoff_value, b=cutoff_value)
```
dirkgr, line 19: Why is this needed?

olmo/model.py
dirkgr: Can you upstream this, so we don't have to depend on a private fork?

olmo/model.py
```diff
@@ -660,10 +660,245 @@ def build(cls, layer_id: int, config: ModelConfig, cache: BufferCache) -> OLMoBl
...
+                x = self._activation_checkpoint_fn(self.ff_norm, x)  # type: ignore
+            else:
+                x = self.ff_norm(x)
+            return og_x + self.dropout(x), cache
+        else:
+            if self._activation_checkpoint_fn is not None:
+                x = self._activation_checkpoint_fn(self.ff_norm, x)  # type: ignore
+            else:
+                x = self.ff_norm(x)
+            # Activation checkpointing for the MoE FFN is not supported
```
dirkgr, line 898: Why not? If there is a technical problem with it, will it affect `whole_layer` activation checkpointing as well?

olmo/model.py
```diff
@@ -1538,7 +1780,7 @@ def fsdp_wrap_fn(module, recurse: bool = True, nonwrapped_numel: int = 0):
         else:
             raise NotImplementedError(wrap_strategy)
 
-    def num_params(self, include_embedding: bool = True) -> int:
+    def num_params(self, include_embedding: bool = True, include_inactivate_params: bool = True) -> int:
```
dirkgr, line 1783: What is "inactivate"?

olmo/train.py
```diff
@@ -237,6 +259,31 @@ def __post_init__(self):
...
+                "device": self.model.config.init_device,
+                "moe_capacity_factor": self.model.config.moe_capacity_factor,
+                "moe_loss_weight": self.model.config.moe_loss_weight,
+                "fp16": False,
+                "bf16": False,
+                "shared_expert": self.model.config.moe_shared_expert,
+                "moe_lbl_in_fp32": self.model.config.moe_lbl_in_fp32,
+            }
+            if self.model.config.moe_zloss_weight:
+                kwargs["moe_zloss_weight"] = self.model.config.moe_zloss_weight
```
dirkgr, line 283: This whole thing exists twice. Can you make a function that translates an OLMo config to `MoEArgs` and use it in both places? Otherwise they will drift apart.

olmo/train.py
```diff
@@ -727,6 +782,21 @@ def train_batch(self, batch: Dict[str, Any]) -> Tuple[torch.Tensor, Optional[tor
 
         ce_batch_loss = torch.tensor(0.0, device=self.device)
         z_batch_loss = None if not self.cfg.softmax_auxiliary_loss else torch.tensor(0.0, device=self.device)
+        lb_batch_loss = (
+            None if self.model.config.block_type != BlockType.moe else torch.tensor(0.0, device=self.device)
+        )
+        moe_z_batch_loss = (
```
dirkgr, line 788: Why is MoE zloss different from regular zloss? What happens if you enable both?

olmo/train.py
```diff
@@ -753,10 +823,31 @@ def train_batch(self, batch: Dict[str, Any]) -> Tuple[torch.Tensor, Optional[tor
...
+                        if self.model.config.moe_zloss_weight:
+                            tokens_per_expert, _, _ = zip(*get_load_balancing_loss())
+                        else:
+                            tokens_per_expert, _ = zip(*get_load_balancing_loss())
+                        expert_assignments += torch.stack(tokens_per_expert, dim=0).cpu()
+                    clear_load_balancing_loss()
+                    if self.model.config.moe_loss_weight:
+                        loss += lb_loss
+                    if self.model.config.moe_zloss_weight:
+                        loss += moe_z_loss
```
dirkgr, line 843: Where do those weights get applied?

dirkgr: It looks like they are just used as a boolean flag.

scripts/train.py
```diff
@@ -131,7 +131,9 @@ def main(cfg: TrainConfig) -> None:
     olmo_model = OLMo(cfg.model)
     log.info(f"Total number of parameters: {olmo_model.num_params():,d}")
     log.info(f"Number of non-embedding parameters: {olmo_model.num_params(include_embedding=False):,d}")
-    log.info(f"Peak GPU Memory (MB) before {cfg.distributed_strategy}: {int(peak_gpu_memory() or 0)}")
+    if olmo_model.config.block_type == "moe":
+        log.info(f"Number of active parameters: {olmo_model.num_params(include_inactivate_params=False):,d}")
+    log.info(f"Peak GPU Memory (MB) before FSDP: {int(peak_gpu_memory() or 0)}")
```
dirkgr, line 136: I think something went wrong with the merge here. You changed the message.

dirkgr (changes requested):

olmo/config.py
Muennighoff: They're optional when no MoE is used, otherwise required. Is this not an acceptable usage of `Optional[int]`? Can change it

olmo/initialization.py
Muennighoff: Because some weights in the MoE are `torch.nn.Parameter` and do not have a `weight` attribute

olmo/train.py
Muennighoff: In megablocks via the MoE config (`"moe_loss_weight": self.model.config.moe_loss_weight,`)

olmo/model.py
Muennighoff: It fails with

`torch.utils.checkpoint.CheckpointError: torch.utils.checkpoint: Unpack is being triggered for a tensor that was already unpacked once. If you are calling ctx.saved_tensors in backward, make sure to do so only once. Otherwise please open an issue with details on your use case. 2024-05-15T20:15:01.172963498Z 2024-05-15 13:15:01.171	jupiter-cs-aus-133.reviz.ai2.in:3	olmo.util:158	CRITICAL	Uncaught CheckpointError: torch.utils.checkpoint: Unpack is being triggered for a tensor that was already unpacked once. If you are calling ctx.saved_tensors in backward, make sure to do so only once. Otherwise please open an issue with details on your use case.`

This paper has some explanations why it is difficult to do act ckpt for MoEs: https://dspace.mit.edu/bitstream/handle/1721.1/153897/wisdom-dwisdom-meng-eecs-2024-thesis.pdf

`whole_layer` is not supported with MoE, only `fine_grained` - I added code to raise an error if it's not `fine_grained` & MoE is configured.

olmo/model.py
Muennighoff: Fixed to `inactive`

scripts/train.py
Muennighoff: Fixed!

olmo/train.py
Muennighoff: Done!

olmo/train.py
Muennighoff: It's explained in this paper: https://arxiv.org/pdf/2202.08906 - essentially the MoE one is applied to routers
You can activate both if you want though I haven't tried it

olmo/model.py
Muennighoff: Sure, opened a PR here https://github.com/databricks/megablocks/pull/133 - If / when it gets merged, I will update the install instructions. If people don't want to use zloss, it also works with the regular megablocks - it's not a big difference.

Muennighoff:
Linking this related PR that we should merge after: https://github.com/allenai/OLMo/pull/707

If this PR here looks good to you, could you approve it @epwalsh / @dirkgr ? :)

olmo/model.py
```diff
@@ -662,10 +662,219 @@ def build(cls, layer_id: int, config: ModelConfig, cache: BufferCache) -> OLMoBl
...
+        try:
+            from megablocks.layers.dmoe import dMoE
+            from megablocks.layers.moe import MoE
+        except ImportError:
+            raise ImportError(
+                "To train MoEs, run `pip install git+https://github.com/Muennighoff/megablocks.git@olmoe`"
+            )
+        from .config import config_to_moe_args
+
+        nn.Module.__init__(self)
```
2015aroras, line 687: This constructor seems a bit hacky with the way it tries to make `OLMoEBlock` implement `OLMoBlock` but not use the parent constructor. I think it would be better to, for example, change the `OLMoBlock` constructor so it has the common functionality between all types of blocks and move the block-specific stuff into each block's constructor.

olmo/optim.py
```diff
@@ -858,6 +865,12 @@ def get_param_groups(cfg: TrainConfig, model: nn.Module) -> List[Dict[str, Any]]
                     decay.add(fpn)
                 else:
                     no_decay.add(fpn)
+            elif (
+                megablocks_available
+                and pn.endswith(("w1", "w2", "v1"))
+                and (isinstance(m, MLP) or isinstance(m, SparseMLP))
```
2015aroras, line 871: nit
```suggestion
                and isinstance(m, (MLP, SparseMLP))
```

configs/official/OLMoE-7B-A1B.yaml
```diff
@@ -0,0 +1,1494 @@
...
+  t_warmup: 10485760000
+  t_max: 5e12
+  alpha_f: 0.1
+
+tokenizer:
+  identifier: tokenizers/allenai_gpt-neox-olmo-dolma-v1_5.json
+  truncate_direction: right
+
+save_folder: runs/${run_name}
+remote_save_folder: s3://ai2-llm/checkpoints/OLMoE/${run_name}
```
2015aroras, line 73: nit: remove this

configs/official/OLMoE-7B-A1B.yaml
```diff
@@ -0,0 +1,1494 @@
+run_name: ${oc.env:SLURM_JOB_ID}
```
2015aroras, line 1: nit: change this to some constant not depending on the environment

olmo/config.py
Muennighoff: I can change it to option 1) if others agree? Note that there's other params not following this:
```
    embedding_size: Optional[int] = 50304
    gen1_gc_interval: Optional[int] = 1
    distributed_strategy: Optional[DistributedStrategy] = DistributedStrategy.fsdp
    fsdp: Optional[FSDPConfig] = field(default_factory=FSDPConfig)
    auxiliary_loss_multiplier: Optional[float] = 1e-4
```

olmo/config.py
```diff
@@ -1273,3 +1334,41 @@ def update_legacy_settings(cls, config: D) -> D:
                 new_config.optimizer = OptimizerConfig.update_legacy_settings(new_config.optimizer)
 
         return new_config
+
+
+def config_to_moe_args(config: ModelConfig) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
Muennighoff: I think the moe args may include things outside of the ModelConfig in the future. Currently, I put some things that may be considered as TrainingConfig params like moe_zloss_weight in the ModelConfig but in case we move them in the future to TrainingConfig then it would not only use the ModelConfig anymore.

olmo/model.py
Muennighoff: Changed 👍

Muennighoff:
All tests are passing except the GPU test which I assume is expected to fail. Feel free to merge 😊

olmo/config.py
dirkgr: Do you actually rely on the defaults you put in here anywhere? If not, let's go with Shane's version, and default these to `None`. I assume something somewhere will fail if they are not set and you need them.

Muennighoff: > Do you actually rely on the defaults you put in here anywhere?

Yes quite a lot, e.g. the loss weights; the use of dropless MoEs (moe_dropless); leaving moe_interleave,moe_lbl_in_fp32,moe_shared_expert as False

Actually, I don't think setting them all to None is a good idea, as it means that everytime we add a new MoE-specific configuration parameter all MoE configs become outdated since every MoE-specific configuration parameter is Optional in that dense.

I can also remove the `Optional` from it as they have defaults anyways but then as seen in the examples I pasted above, we do have `Optional` config params with default values in the codebase anyways.

olmo/model.py
```diff
@@ -448,14 +448,15 @@ def __init__(self, layer_id: int, config: ModelConfig, cache: BufferCache):
...
 
-        # Feed-forward output projection.
-        self.ff_out = nn.Linear(
-            int(self.act.output_multiplier * self.hidden_size),
-            config.d_model,
-            bias=config.include_bias,
-            device=config.init_device,
-        )
-        self.ff_out._is_residual = True  # type: ignore
+        if self.config.block_type != BlockType.moe:
```
dirkgr, line 451: Can you make this dependent on whether the block has a `ff_out`, instead of the block type?

olmo/model.py
dirkgr: @Muennighoff , so they decided to merge their version instead. Is our version compatible? Will the model you trained work with their implementation of zloss?

olmo/model.py
dirkgr: with `hasattr()`, I mean

olmo/model.py
```diff
@@ -662,10 +663,164 @@ def build(cls, layer_id: int, config: ModelConfig, cache: BufferCache) -> OLMoBl
...
+            raise NotImplementedError(self.config.init_fn)
+
+        init_normal(self.att_proj, std=in_std, init_cutoff_factor=cutoff_factor)
+        init_normal(self.attn_out, std=attn_out_std, init_cutoff_factor=cutoff_factor)
+        self.attn_norm.reset_parameters()
+        self.ff_norm.reset_parameters()
+        init_normal(self.ffn.experts.mlp.w1, std=in_std, init_cutoff_factor=cutoff_factor)
+        init_normal(self.ffn.experts.mlp.w2, std=ff_out_std, init_cutoff_factor=cutoff_factor)
+        if hasattr(self.ffn.experts.mlp, "v1"):
+            init_normal(self.ffn.experts.mlp.v1, std=in_std, init_cutoff_factor=cutoff_factor)
```
dirkgr, line 735: What is `v1` and why is it optional? Maybe I'll keep reading and find out.

olmo/model.py
dirkgr: Ok, I see. Interesting. It would be fixable I think (by saving the active experts per token in the forward pass), but out of scope for this PR.

dirkgr: This is probably a fairly big blocker to going bigger though. For dense models, our fastest settings still use a lot of checkpointing.

olmo/model.py
```diff
@@ -1100,8 +1255,15 @@ def __init__(self, config: ModelConfig, init_params: bool = True):
...
         )
-
-        blocks = [OLMoBlock.build(i, config, self.__cache) for i in range(config.n_layers)]
+        if self.config.moe_interleave:
+            blocks = []
+            for i in range(config.n_layers):
+                if i % 2 == 0:
+                    blocks.append(OLMoSequentialBlock(i, config, self.__cache))
+                else:
+                    blocks.append(OLMoEBlock(i, config, self.__cache))
```
dirkgr, line 1264: If we don't need this, I'd rather not have it.

olmo/train.py
```diff
@@ -777,6 +811,21 @@ def train_batch(self, batch: Dict[str, Any]) -> Tuple[torch.Tensor, Optional[tor
...
+            None if not self.model.config.moe_zloss_weight else torch.tensor(0.0, device=self.device)
+        )
+        # Keep this one on CPU to save memory
+        expert_assignments = (
+            None
+            if (
+                (self.model.config.block_type != BlockType.moe)
+                or (self.model.config.moe_log_expert_assignment is False)
+            )
+            else torch.zeros((self.model.config.n_layers, self.model.config.moe_num_experts))
```
dirkgr, line 827: This does not put it on CPU. This puts it on the default device.

olmo/train.py
```diff
@@ -807,14 +856,35 @@ def train_batch(self, batch: Dict[str, Any]) -> Tuple[torch.Tensor, Optional[tor
...
+
                 # Run backward pass.
                 loss.backward()
 
             # Remove output hooks
             for hook in output_hooks:
                 hook.remove()
 
-        return ce_batch_loss, z_batch_loss
+        return ce_batch_loss, z_batch_loss, lb_batch_loss, moe_z_batch_loss, expert_assignments
```
dirkgr, line 887: @epwalsh, does the new trainer support all of this stuff? This seems like a lot of extra things.

olmo/train.py
```diff
@@ -807,14 +856,35 @@ def train_batch(self, batch: Dict[str, Any]) -> Tuple[torch.Tensor, Optional[tor
...
+                if self.model.config.block_type == BlockType.moe:
+                    if self.model.config.moe_zloss_weight is not None:
+                        lb_loss, moe_z_loss = batched_load_balancing_loss(self.moe_args)
+                        lb_loss = lb_loss / len(micro_batches)
+                        moe_z_loss = moe_z_loss / len(micro_batches)
+                    else:
+                        lb_loss = batched_load_balancing_loss(self.moe_args) / len(micro_batches)
+                    if self.model.config.moe_log_expert_assignment:
+                        if self.model.config.moe_zloss_weight:
+                            tokens_per_expert, _, _ = zip(*get_load_balancing_loss())
```
dirkgr, line 868: Goes `get_load_balancing_loss()` take care of reducing the expert assignments across ranks?

olmo/train.py
```diff
@@ -807,14 +856,35 @@ def train_batch(self, batch: Dict[str, Any]) -> Tuple[torch.Tensor, Optional[tor
...
+                        else:
+                            tokens_per_expert, _ = zip(*get_load_balancing_loss())
+                        expert_assignments += torch.stack(tokens_per_expert, dim=0).cpu()
```
dirkgr, line 871: Not a big deal, but are you sure that this is faster on CPU? Back in the day I always thought this too, keep small stuff on the CPU, but in practice doing it all on GPU was always faster.

dirkgr (changes requested):

olmo/config.py
dirkgr: If it doesn't break everything, I'd prefer to have a special config object for MoE, which is `Optional`, but none of the items inside of that object are `Optional`. This may break backwards compatibility with the model we already released though?

olmo/train.py
epwalsh: Not directly but I think it could be supported through the callback system.

olmo/train.py
epwalsh: If `tokens_per_expert` were on GPU this will trigger a host-device sync. In that case it's almost certainly better to keep on GPU.

Muennighoff: > were on GPU this will trigger a host-device sync

Don't we want to avoid these syncs?

olmo/config.py
Muennighoff: Yes it would break compat with the configs we released but can pin a commit to our released repo if people want to reuse our configs to reproduce things exactly

olmo/model.py
Muennighoff: It is when SwiGLU is activated with MoEs

olmo/config.py
dirkgr: Hm, that's unfortunate, but I think I prefer the `MoEConfigObject`. It reduces the impact on old-school dense model training.

olmo/config.py
```diff
@@ -457,6 +463,61 @@ class ModelConfig(BaseConfig):
...
+
+    moe_lbl_in_fp32: Optional[bool] = False
+    """
+    Whether to perform load balancing in FP32.
+    """
+
+    moe_interleave: Optional[bool] = False
+    """
+    Interleave sequential with MoE blocks starting with sequential.
+    """
```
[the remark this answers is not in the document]
Muennighoff: Removed

olmo/model.py
Muennighoff: Do you mean `if hasattr(self, "ff_out"):`? Not sure that will work because the next lines are about creating self.ff_out so no block has it yet afaict

olmo/train.py
Muennighoff: Moved it to GPU now

olmo/train.py
Muennighoff: changed to place it on self.device

olmo/config.py
Muennighoff: I guess it would make the name ModelConfig a bit inaccurate though; maybe it should inherit from ModelConfig or sth

Muennighoff:
> What's going on with this PR? Can we merge?

Fixed some basics as discussed; I think we can merge!

soldni:
@dirkgr shall we merge?

PR #649, opened 2024-07-10
Model Ladder

dirkgr:
Adds a script that should let us establish scaling properties of interventions we're considering.

TODO:
 - [x] Final eval doesn't get created
 - [x] Run all sizes and fix what breaks.
 - [ ] Add a 3B config?
 - [x] See if using `torch.compile()` is trustworthy.
 - [x] See if we can optimize perf in other ways.
 - [ ] The 300M, 750M, and 1B all have 16 heads and 16 layers. Shouldn’t those vary in scale a little?
 - [x] Make it work on only one node
 - [x] Make it possible to continue runs (i.e., add a `--load_path` flag)
 - [x] Pick a perplexity set to target
 - [ ] Build a perplexity set that's based on the downstream tasks we care about
 - [ ] Find out why the downstream tasks are so random. Logits almost the same?
 - [ ] Find out if there are differences between 1B with FSDP and 1B with DDP


olmo/train.py
```diff
@@ -259,26 +256,8 @@ def max_steps(self) -> int:
...
-            else:
-                # convert to float *first* to handle scientific notation
-                return (
-                    self.global_train_tokens_seen
-                    + max(int(float(self.cfg.max_duration)) - self.global_step, 0) * self.tokens_per_batch
-                )
-        else:
-            raise TypeError(f"expected int or str for 'max_duration', found {type(self.cfg.max_duration)}")
+        # It should be noted that this math rounds up the number of tokens to be a multiple of the batch size.
+        return self.max_steps * self.tokens_per_batch
```
epwalsh, line 260: This doesn't work if batch size has changed at some point

scripts/ladder.py
```diff
@@ -0,0 +1,436 @@
...
+    layer_norm_eps=1e-6,
+    bias_for_layer_norm=False,
+    attention_layer_norm_with_affine=False,
+    activation_type=ActivationType.swiglu,
+    residual_dropout=0.0,
+    embedding_dropout=0.0,
+    max_sequence_length=2048,
+    vocab_size=50280,
+    embedding_size=50304,
+    eos_token_id=0,
```
epwalsh, line 71: This actually be `50279`.
```suggestion
    eos_token_id=50279,
```

scripts/ladder.py
```diff
@@ -0,0 +1,436 @@
...
+        scheduler=SchedulerConfig(
+            name=SchedulerType.cosine_with_warmup,
+            alpha_f=0.01,
+            warmup_min_lr=0.0,
+            t_warmup=round(model_size / (global_batch_size * model_config.max_sequence_length)),
+        ),
+        max_duration=f"{length_in_tokens}T",
+        global_train_batch_size=global_batch_size,
+        tokenizer=TokenizerConfig(identifier="tokenizers/allenai_gpt-neox-olmo-dolma-v1_5.json"),
+        save_folder=f"runs/{run_name}",
```
epwalsh, line 206: Put this on weka?

olmo/data/named_data_mixes.py
```diff
@@ -0,0 +1,1061 @@
+DATA_PATHS = {
+    "dolma17": [
```
epwalsh, line 2: Maybe call this `dolma1_7`?

epwalsh (approved):
LGTM overall, just a couple minor comments

scripts/ladder.py
dirkgr: I tried that first. Didn't work. Weka wasn't writeable. But also, it doesn't matter while the remote save folder is in s3. This folder is only temporary.

epwalsh: Ok but it should be writable, so not sure what's going on there

olmo/data/named_data_mixes.py
dirkgr: I don't like that because it reads like `("dolma1", "7")`. Also, apparently all of our future releases are going to be named like "0727" anyways, so there is no way we'll get into confusion with a possible seventeenth version of Dolma.

scripts/ladder.py
epwalsh: I haven't had an issue with my recent runs. They all write to weka.

scripts/ladder.py
```diff
@@ -0,0 +1,436 @@
...
+    model_size, model_size_unit = _number_unit_re.match(size.strip().upper()).groups()  # type: ignore
+    model_size = int(model_size)
+    if model_size_unit == "K":
+        model_size *= 1000
+    elif model_size_unit == "M":
+        model_size *= 1000000
+    elif model_size_unit == "B":
+        model_size *= 1000000000
+    else:
+        raise ValueError(f"Could not parse model name '{args.model}'")
```
2015aroras, line 103:
```suggestion
        raise ValueError(f"Could not parse model size unit '{model_size_unit}'")
```

scripts/beaker/ladder-launch.sh
```diff
@@ -0,0 +1,39 @@
...
+  COMMAND="scripts/beaker/ladder.sh \$BEAKER_LEADER_REPLICA_HOSTNAME ${NUM_NODES} \$BEAKER_REPLICA_RANK $*"
+fi
+
+gantry run \
+  --workspace ai2/OLMo-training \
+  --task-name ladder \
+  --description "OLMo ladder with $*" \
+  --priority normal \
+  --preemptible \
+  --beaker-image shanea/olmo-torch2.2-gantry \
```
2015aroras, line 22: Consider upgrading this to `petew/olmo-torch23-gantry` or `shanea/olmo-torch23-gantry` (untested but has `importlib_resources`)

2015aroras (approved):

scripts/ladder.py
```diff
@@ -0,0 +1,436 @@
...
+    return length_in_tokens
+
+
+def config_from_args(args: argparse.Namespace) -> TrainConfig:
+    # Construct a config
+    args.model = args.model.strip().upper()
+    run_name = f"{args.name}-{args.model}-{args.length}"
+    assert "/" not in run_name
+
+    permanent_data_prefix = "/weka/oe-training-default/ai2-llm"
```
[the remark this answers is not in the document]
dirkgr: Done in cc7c18473ec0b65fc0d31f1a63a0ad7f4485d4d3

scripts/beaker/ladder-launch.sh
dirkgr: Done in c295ceae2a0777fab5e07476dbee1b71a6f2fa4f, though I have a vague memory that I tried this and there was a problem with it.

PR #653, opened 2024-07-11
Bump version to v0.4.0

2015aroras:


epwalsh (commented):
Wait why wasn't this handled by the release script?

epwalsh:
Ah, permissions issue I guess? I'll fix your permissions for future releases.

epwalsh (approved):

olmo/version.py
```diff
@@ -1,5 +1,5 @@
 _MAJOR = "0"
-_MINOR = "3"
+_MINOR = "4"
 # On main and in a nightly release the patch should be one ahead of the last
 # released build.
 _PATCH = "1"
```
epwalsh, line 5:
```suggestion
_PATCH = "0"
```

2015aroras: I thought we kept the repo at version + 1 (so v0.4.1 when we are at 0.4.0)

2015aroras:
I ran the release script but I couldn't do it from main (maybe permissions)? It's not clear whether to do it from main or a branch

olmo/version.py
epwalsh: No we typically leave it as is until the next release

epwalsh:
> I ran the release script but I couldn't do it from main (maybe permissions)? It's not clear whether to do it from main or a branch

You should do it from main. I fixed your permissions so next time it should work for you.

PR #659, opened 2024-07-16
Add option not to remove tmp directory when doing `convert_olmo_to_hf_new.py`

dwadden:
Attempting to remove it causes an error when using NFS.

### DIFF
--- a/scripts/s3_unshard_to_hf.py
+++ b/scripts/s3_unshard_to_hf.py
@@ -8,7 +8,7 @@ python scripts/s3_unshard_to_hf.py \
     --unsharded_bucket s3://ai2-llm/checkpoints/OLMo-medium/mitchish7/step239000-unsharded \
     --hf_bucket s3://ai2-llm/checkpoints/OLMo-medium/mitchish7/step239000-huggingface \
     --type olmo_core \
-    --tmp_dir /net/nfs.cirrascale/allennlp/davidw/tmp/unshard
+    --local_dir /net/nfs.cirrascale/allennlp/davidw/tmp/unshard
 
 NOTE: For this to work, you need to install the `OLMo-core` repo as follows:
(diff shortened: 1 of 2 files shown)


scripts/s3_unshard_to_hf.py
```diff
@@ -31,91 +31,126 @@ def make_parser():
...
+            subprocess.run(download_cmd, shell=True, check=True)
+
+        # Unshard.
+        print("Unsharding.")
+        unshard_cmd = f"python scripts/unshard.py {sharded_dir} {unsharded_dir}"
+        # Add a `--type` argument if given.
+        if args.type is not None:
+            unshard_cmd += f" --type {args.type}"
+        if args.model_only:
+            unshard_cmd += " --model-only"
```
2015aroras, line 105: nit: `model_only`

scripts/s3_unshard_to_hf.py
2015aroras, line 1: I'm not sure this file is supposed to be in this PR? Or at least only the part with `tmp_cleanup` is supposed to be?

dwadden: Good point, I can make a separate PR for this file. I'll tag you to review if you don't mind.

scripts/s3_unshard_to_hf.py
dwadden: I was matching the hyphen in [unshard.py](https://github.com/allenai/OLMo/blob/main/scripts/unshard.py#L90), is this incorrect practice?

dwadden:
I'll split this out into 2 separate PR's and do again.

scripts/s3_unshard_to_hf.py
2015aroras: My bad, I saw you change `model-only` to `model_only` a few lines earlier so I commented without thinking much about it

PR #661, opened 2024-07-16
Add support for document masking during training

epwalsh:
Adds support for [document masking](https://www.semanticscholar.org/paper/c96b17f3362c2d795964d07427a9396a09db6b2c) during training via flash-attn.
This is activated when the flag `--data.generate_doc_lengths` is set.
The code changes were adapted from https://github.com/yuzhaouoe/pretraining-data-packing.


AkshitaB (approved):

dirkgr (approved):
If this is a good change, we definitely need to use flash attention on LUMI, which is its own can of worms ...

PR #663, opened 2024-07-17
Improving memmap type parser

soldni:
The old `effective_memmap_dtype` property will default to `uint16` in case the dtype provided is not valid. This is bad because, as @drschwenk found, value `numpy.uint32` will cause `uint16` to be used. 

New property will raise an `TypeError` if dtype is not recognized.


drschwenk (approved):
LGTM!

dirkgr (approved):
It's broader than I'd have done it, but that's OK. Not a big fan of silent errors. Massive fan of quick fixes.

PR #670, opened 2024-07-18
Fix restarts in later epochs

epwalsh:
This PR fixes restarting a training run in later epochs so that we no longer need to set the flag `--epoch=INT`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed default distributed training strategy from single-GPU to FSDP
 - Fixed behavior of `effective_memmap_dtype` to prevent unrecognized dtypes to be parsed as `uint16`. 
 
+### Fixed
+
+- Fixed restarting a training run in later epochs so that we no longer need to set the flag `--epoch=INT`. 
+
 ## [v0.4.0](https://github.com/allenai/OLMo/releases/tag/v0.4.0) - 2024-07-11
 
 ### Added
--- a/olmo/train.py
+++ b/olmo/train.py
@@ -1269,7 +1269,7 @@ class Trainer:
                             python_profiler.print_stats(sort=SortKey.CUMULATIVE)
                             python_profiler = None
                 else:
-                    log.info(f"Epoch {epoch} complete")
+                    log.info("Training epoch complete")
                     self.epoch = epoch + 1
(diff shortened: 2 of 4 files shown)


olmo/train.py
```diff
@@ -352,7 +352,7 @@ def load_trainer_state_dict(self, state_dict: Dict[str, Any]) -> None:
         ]
 
         # Dataset / dataloader position.
-        checkpoint_epoch = state_dict.get("epoch", 0)
+        checkpoint_epoch = state_dict.get("epoch") or 0
```
Muennighoff, line 355: Now it will error out if the state_dict does not have the `epoch` key right? Will it always be present going forward?

olmo/train.py
```diff
@@ -1263,12 +1269,13 @@ def on_trace_ready(p):
                             python_profiler.print_stats(sort=SortKey.CUMULATIVE)
                             python_profiler = None
                 else:
-                    log.info("Training epoch complete")
+                    log.info(f"Epoch {epoch} complete")
```
Muennighoff, line 1272: One thing that may be confusing is that it will say `Epoch 0 complete` and `Epoch 1 complete` will actually mean that 2 epochs have been completed. But fine to leave as is I think.

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed default distributed training strategy from single-GPU to FSDP
 - Fixed behavior of `effective_memmap_dtype` to prevent unrecognized dtypes to be parsed as `uint16`. 
 
+### Fixed
+
+- Fixed restarting a training run in later epochs so that we no longer need to set the flag `--epoch=INT`.
```
Muennighoff, line 22: Should we remove the `--epoch` flag?

olmo/train.py
epwalsh: `.get()` won't throw a `KeyError` so I think it's okay

CHANGELOG.md
epwalsh: It might be useful at some point so I'd rather keep it for now

olmo/train.py
epwalsh: yea true. Actually since we're logging the epoch along with step number I think I'll just revert this.

epwalsh: 4c7ada39

Muennighoff (approved):

PR #680, opened 2024-07-30
Ensure attn_norm does not affect the residual stream

2015aroras:
Issue: The attention norm, when applied before the attention block, was modifying the residual stream. This was introduced in https://github.com/allenai/OLMo/commit/aac4bdd1b417c379b79fdba225bbbe01cac9e4ef.

Fix: Ensure the attention norm does not modify the residual stream.

Testing: I used a script to check that this PR made the output match that from https://github.com/allenai/OLMo/tree/ladder-1xC, which is healthy according to @AkshitaB. Before the fix, the outputs (and inputs) would start disagreeing at the feed forward norm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


epwalsh (approved):
Good catch.

PR #671, opened 2024-07-19
Unshard S3 checkpoints and convert to HF.

dwadden:
Updates to script to handle unsharding and HF conversion of checkpoints stored on S3.

PR #684, opened 2024-08-01
Revamped HF conversion with support for unsharding, upload/download from s3. 

soldni:


soldni:
ah @dwadden this seems to be duplicated functionality from #671 ?

PR #687, opened 2024-08-02
Extend functionality of Wandb Config Diff script

kyleclo:
- Add tests for `flatten_dict()` in utils
- Extend functionality for `flatten_dict()` to also flatten any dicts that exist in Lists
- Extend the wandb config comparison script to use the extended `flatten_dict()`

Motivation is, while comparing configs, the current implementation doesn't perform comparison of some key aspects of the configs, namely config keys representing dataset paths (which are all List[str]) as well as keys like `config["evaluators.value"]` which are `List[Dict]`.

The current behavior looks something like this:
image

where we can see that the fields `data.value.paths` and `evaluators.value` aren't easily comparable.

The new behavior looks like this:
image
where it preserves behavior of original script under old keys, but performs a specialized diff just for `data.paths` and `evaluators`.  

for `data.paths`, the idea is that we just need to know the names of the datasets from the paths + how many files used.  Added new function that does that specifically.


olmo/util.py
```diff
@@ -841,12 +841,28 @@ def get_bytes_range(self, index: int, length: int) -> bytes:
...
-def flatten_dict(dictionary, parent_key="", separator="."):
+def flatten_dict(dictionary, parent_key="", separator=".", include_lists=False):
+    """
+    Flatten a nested dictionary into a single-level dictionary.
+
+    Args:
+        dictionary (dict): The nested dictionary to be flattened.
+        parent_key (str, optional): The parent key to be prepended to the keys of the flattened dictionary. Defaults to "".
+        separator (str, optional): The separator to be used between the parent key and the keys of the flattened dictionary. Defaults to ".".
+        include_lists (bool, optional): Whether to convert lists to dictionaries with integer keys. Defaults to False.
```
dirkgr, line 852: Do we ever want to turn this off?

scripts/compare_wandb_configs.py
```diff
@@ -24,6 +24,47 @@ def parse_run_path(run_path: str) -> str:
...
+        s_right += prefix + "Settings only in right:\n"
+        s_right += (prefix + "\n").join(f"\t{k}: {right_config[k]}" for k in sorted(right_only_keys)) + "\n"
+
+    s_shared = ""
+    keys_with_differences = {
+        k for k in left_config.keys() & right_config.keys() if left_config[k] != right_config[k]
+    }
+    if len(keys_with_differences) > 0:
+        for k in sorted(keys_with_differences):
+            if isinstance(left_config[k], list) and isinstance(right_config[k], list):
```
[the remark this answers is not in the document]
dirkgr: Can you simplify this now that lists are properly flattened?

scripts/compare_wandb_configs.py
```diff
@@ -24,6 +38,78 @@ def parse_run_path(run_path: str) -> str:
...
+
+def simplify_path(path):
+    # Remove the variable prefix up to and including 'preprocessed/'
+    path = re.sub(r"^.*?preprocessed/", "", path)
+
+    # Split the path into components
+    components = path.split("/")
+
+    # Remove common suffixes like 'allenai'
+    components = [c for c in components if c != "allenai"]
```
dirkgr, line 90: You don't think that's confusing? For copy and paste?

scripts/compare_wandb_configs.py
```diff
@@ -24,6 +38,78 @@ def parse_run_path(run_path: str) -> str:
...
+    for path, num_files in simplified_left_data_paths.items():
+        print(f"\t{path}: {num_files}")
+    print("\n\n")
+
+    print("===== Data Paths for right config:\n")
+    simplified_right_data_paths = {simplify_path(path): count for path, count in right_data_paths.items()}
+    for path, num_files in simplified_right_data_paths.items():
+        print(f"\t{path}: {num_files}")
+
+    return
```
dirkgr, line 110:
```suggestion
```

scripts/compare_wandb_configs.py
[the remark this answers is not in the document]
dirkgr: Actually, it seems this will never happen now?

dirkgr (changes requested):

dirkgr:
You let me know when this is ready for another look?

olmo/util.py
kyleclo: not for now, but seems fine to extend this in case someone wants to add different logic for dealing w list-valued config params?

scripts/compare_wandb_configs.py
kyleclo: removed

dirkgr (approved):

PR #690, opened 2024-08-05
Add ability to save and compare sub-module outputs

2015aroras:
This PR adds the ability to save and compare sub-module outputs. This was helpful for finding the fix in #680. We could use this functionality as part of an integration test in the future.

### DIFF
--- a/olmo/config.py
+++ b/olmo/config.py
@@ -1219,7 +1219,8 @@ class TrainConfig(BaseConfig):
 
     module_outputs_save_steps: Optional[List[int]] = None
     """
-    Outputs of model submodules are saved during the provided steps.
+    Outputs of model submodules are saved during the provided steps. Submodule outputs
+    can be compared using `scripts/compare_module_outputs.py`.
     """
 
     @property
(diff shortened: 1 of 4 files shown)


olmo/config.py
```diff
@@ -1217,6 +1217,11 @@ class TrainConfig(BaseConfig):
     Path to cache directory of HF datasets saved with `datasets.save_to_disk`.
     """
 
+    module_outputs_save_steps: Optional[List[int]] = None
+    """
+    Outputs of model submodules are saved during the provided steps.
```
epwalsh, line 1222: Can you add a note here about how to use the script to view the results?

epwalsh (approved):
LGTM

olmo/config.py
2015aroras: https://github.com/allenai/OLMo/pull/690/commits/81c07c6c75210c25d7255b7e5665a7076d04027e I've mentioned the script for now. I intend to add a doc describing debugging with this soon.

PR #691, opened 2024-08-06
Add scheduler for cosine in linear envelope.

dwadden:
This didn't prove particularly effective for annealing, but might as well have it as an option.

### DIFF
--- a/olmo/optim.py
+++ b/olmo/optim.py
@@ -792,17 +792,24 @@ class ConstantScheduler(Scheduler):
 @dataclass
 class CosLinearEnvelope(Scheduler):
     "Pointwise product of cosine schedule and linear decay; useful during annealing."
+    warmup_steps: int
+    alpha_f: float = 0.1
     t_max: Optional[int] = None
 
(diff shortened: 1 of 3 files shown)


olmo/optim.py
```diff
@@ -788,6 +789,22 @@ def get_lr(self, initial_lr: float, step: int, max_steps: int) -> float:
         return initial_lr
 
 
+@dataclass
+class CosLinearEnvelope(Scheduler):
+    "Pointwise product of cosine schedule and linear decay; useful during annealing."
+    t_max: Optional[int] = None
+
+    def get_lr(self, initial_lr: float, step: int, max_steps: int) -> float:
+        eta_min = 0.0
```
2015aroras, line 798: Consider setting this to `initial_lr * alpha_f`, like with the cosine and linear schedulers. Warm up would probably be nice too.

2015aroras: If you use `alpha_f`, you may need to update the logic for computing the envelope.

dwadden: Changed, lmk if this looks good.

2015aroras (approved):
LGTM!

PR #698, opened 2024-08-12
Add script for comparing model state

2015aroras:
This PR adds a script for comparing model state of 2 different OLMo models/checkpoints. This helped reveal that one of the LUMI containers was not loading/saving checkpoints correctly.


epwalsh (approved):
Can you add a module docstring to the top of the script explaining what it does exactly? Otherwise LGTM.

> This helped reveal that one of the LUMI containers was not loading/saving checkpoints correctly.

Did you get to the bottom of this? What exactly was the issue and the fix?

2015aroras:
> > This helped reveal that one of the LUMI containers was not loading/saving checkpoints correctly.
> 
> Did you get to the bottom of this? What exactly was the issue and the fix?

Honestly, no idea what was causing the issue. The issue was in LUMI's provided EasyBuild container but not in any of the containers we made, so I just scrapped the idea of using LUMI's provided container.

PR #705, opened 2024-08-16
Change naming convention for HF checkpoints.

dwadden:
Use `hf_olmo` instead of `old_style` and `transformers` for `new_style`. Add docstring explaining the difference and default to using `hf_olmo`, which is standard for internal development.

I've run this and it seems to work fine.

### DIFF
--- a/scripts/s3_unshard_to_hf.py
+++ b/scripts/s3_unshard_to_hf.py
@@ -53,7 +53,18 @@ def make_parser():
         help="If given, remove the local directory if everything runs successfully to free up space on NFS.",
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


scripts/s3_unshard_to_hf.py
```diff
@@ -111,7 +124,7 @@ def s3_unshard_to_hf(args):
 
     # Convert to HF.
     print("Converting to HF.")
-    if args.old_style_hf:
+    if args.checkpoint_style == "hf_olmo":
         # Convert to old-style checkpoint.
```
2015aroras, line 128: nit: update the comment

2015aroras (approved):
LGTM, thanks!

Issue #683, opened 2024-07-31
Model ladder has no documentation


IanMagnusson:
### 📚 The doc issue

It would be great if there was some kind of a readme for the model ladder (https://github.com/allenai/OLMo/pull/649). As far as I know there is currently no documentation for this.

I believe this is the entry point? https://github.com/allenai/OLMo/blob/main/scripts/ladder.py

@taidnguyen, @dwadden, and I will likely want to run data mix experiments in the future that make use of the model ladder for standardization and having some basic documentation would avoid having each of us have to come bug people for 1:1 explanations.

IanMagnusson:
Intending to address this with https://github.com/allenai/OLMo/pull/708

PR #708, opened 2024-08-19
Docs model ladder

IanMagnusson:
Trying to add some documentation for the model ladder (per https://github.com/allenai/OLMo/issues/683) as I figure out how to use it.


IanMagnusson:
Ideally we don't have to use `--s3` for the usage example, but weka seems to be broken for a number of reasons right now.
1. the current dolma data mix with the flan with rules based sep is not in weka
2. If I revert to an earlier version of the ladder that didn't have this version of flan it still [crashes](https://beaker.org/ex/01J5P705Y89ZVWDZ55XF9NBY2P/tasks/01J5P705Z80KJKX62XMMNPTB0F/job/01J5P70A5CNPS5455VJPKV68KT) without finding training data in weka although that data is there when I check manually.

IanMagnusson:
Would be nice to also have the `-h` response for the other commands in the script (is there an easier way to get these than to do a gantry command to run this with the env all set up. When I try to run it locally I get env issues even tho I have OLMo pip instal -e

Likewise it would maybe be nice to list currently available sizes, lengths, and data mixes.

PR #711, opened 2024-08-22
fix unbound qkv

epwalsh:


chrisc36 (approved):
Look fines. A non-essential nitpick is that variable is still a hidden state, not `qkv`, so I would have preference to rename it as well.

PR #717, opened 2024-08-30
Added ability to try loading the latest checkpoint from save folders

2015aroras:
Issue: Our training runs never finish in 1 run of `train.py`. Currently we don't have a nice way to continue training using the same config; we have to set the `load_path` to the latest checkpoint.

Fix: Add an option that tries loading the latest checkpoint from the local and remote save folders (assuming `load_path` is not set). If there are no checkpoints in either folder, then the model initializes from scratch as usual. If this option (`--try_load_latest_save`) is set to `True` for both the initial and subsequent runs, then the first run will initialize and save an initial checkpoint while subsequent runs will resume from the latest checkpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added ability to try loading latest checkpoint from save folder using `--try_load_latest_save`.
+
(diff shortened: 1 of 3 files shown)


2015aroras:
Tested the 3 main scenarios: no existing checkpoint, only remote checkpoints, local + remote checkpoints.

epwalsh (approved):
LGTM, this will be super useful

PR #718, opened 2024-09-06
Fix mmlu bpb bug only scoring answer=A questions

OyvindTafjord:
Same problem as fixed for oe-eval tasks in https://github.com/allenai/OLMo/pull/712, I forgot there was separate handling for MMLU. 

With the previous code, only questions with gold answer A would be counted in the bpb evaluations, now they should all be counted.

### DIFF
--- a/olmo/eval/downstream.py
+++ b/olmo/eval/downstream.py
@@ -1365,6 +1365,9 @@ class MMLU(ICLMultiChoiceTaskDataset):
             return choices
 
     def doc_to_label(self, doc):
+        if self.metric_type in ["ce_loss", "bpb"]:
+            # Only the correct answer is provided for these metrics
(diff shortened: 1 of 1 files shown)


liujch1998:
Thanks Oyvind! Approving this PR.

Though it seems to me that not resetting label to 0 is fine for MMLU — MMLU’s prep_examples(), which is inherited from ICLMultiChoiceTaskDataset, does not skip cases where label_id and cont_id mismatch when metric=bpb (whereas OEEvalTask does), and ICLMetric.compute() does not use label_id when metric=bpb. So questions with any label_id should have already been included in the metric computation.

liujch1998 (approved):

epwalsh (approved):

PR #721, opened 2024-09-19
Bump torch version

vwxyzjn:


pyproject.toml
```diff
@@ -14,7 +14,7 @@ requires-python = ">=3.8"
 license = { file = "LICENSE" }
 dependencies = [
     "numpy",
-    "torch>=2.1,<2.4",
+    "torch>=2.4",
```
2015aroras, line 17: I don't think we should disallow earlier torch versions.
```suggestion
     "torch>=2.1,<2.5",
```

vwxyzjn:
Good point! Have just changed it.

2015aroras (approved):

PR #731, opened 2024-10-08
Adding script for processing many intermediate checkpoints at once for offline evals

IanMagnusson:
Making a draft PR for this so we can consider merging this in to main. It would be nice if we could do this so we don't run into version issues if we train models in the future that are not compatible with the version of the code forked here.


.gitignore
```diff
@@ -1,3 +1,6 @@
+# beaker yaml
+guided-trout-2f805b9.yaml
```
dirkgr, line 2: What is this?

log.txt
```diff
@@ -0,0 +1,10 @@
+
```
dirkgr, line 1: Don't commit temp output.

hf_olmo/convert_olmo_to_hf.py
```diff
@@ -308,6 +314,9 @@ def main():
     upload_local_checkpoint(local_checkpoint_dir, args.destination_dir)
 
     print(f"Converted checkpoint saved to {args.destination_dir}")
+    if args.cleanup_local_dir:
```
dirkgr, line 317: Is there ever a reason not to do this?

requirements.txt
```diff
@@ -0,0 +1,7 @@
+torch
```
dirkgr, line 1: We don't use requirements.txt in OLMo. We use pyproject.toml.

dirkgr (changes requested):
@soldni , don't you already have a checkpoint converter script that runs in Beaker?

.gitignore
jenahwang: err... i didn't add that? it must be a gantry beaker yaml file based on the naming scheme.

requirements.txt
jenahwang: Oh woops. I added this for gantry. I'll change locations on this.

log.txt
jenahwang: Where is this located? I thought I removed all the temp files I created

.gitignore
jenahwang: i'll remove in any case

hf_olmo/convert_olmo_to_hf.py
jenahwang: i guess not. But it looks like the script has changed since I l pr'd this. So let me make sure my script works with it.

requirements.txt
jenahwang: Removed --- it was created to troubleshoot.

hf_olmo/convert_olmo_to_hf.py
jenahwang: I removed the if statement.

log.txt
jenahwang: sorry. removed

.gitignore
jenahwang: removed

jenahwang:
> @soldni , don't you already have a checkpoint converter script that runs in Beaker?

He does. What this one does is very similar but the focus is on batch conversion and wildcard acceptance. And it was written for oe-eval consistent ranking project with expediting its pipeline in mind.

soldni:
@jenahwang would it be possible to merge in your changes to the other script? or consolidate the two? 

it's confusing to have two conversion scripts, and it doubles maintenance.

PR #736, opened 2024-10-20
Create an eval-only script for existing ckpts

liujch1998:
This PR adds `scripts/eval.py`, which evaluates one or more existing ckpts while bypassing the training steps.

It seems impossible to backfill evals back to the original wandb run, because "step" must always increase. Therefore, this script logs things to a new wandb run.

To run this script, you can specify a single ckpt in as `--load_path`, or all ckpts under a directory if you remove the `/stepXXX` part.

### DIFF
--- a/configs/peteish1-weka.yaml
+++ b/configs/peteish1-weka.yaml
@@ -108,35 +108,35 @@ eval_interval: 1000
 eval_subset_num_batches: -1
 device_eval_batch_size: ${device_train_microbatch_size}
 evaluators:
-  # - label: all-small-ppl-validation
-  #   data:
-  #     num_workers: 0
-  #     drop_last: true
-  #     # generate_doc_lengths: true
-  #     memmap_dtype: uint32
-  #     datasets:
-  #       c4_en-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/c4_en/val/part-0-00000.npy
-  #       dolma_books-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/dolma_books/val/part-0-00000.npy
-  #       dolma_common-crawl-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/dolma_common-crawl/val/part-0-00000.npy
-  #       dolma_pes2o-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/dolma_pes2o/val/part-0-00000.npy
-  #       dolma_reddit-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/dolma_reddit/val/part-0-00000.npy
-  #       dolma_stack-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/dolma_stack/val/part-0-00000.npy
-  #       dolma_wiki-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/dolma_wiki/val/part-0-00000.npy
-  #       ice-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/ice/val/part-0-00000.npy
-  #       m2d2_s2orc-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/m2d2_s2orc/val/part-0-00000.npy
-  #       pile-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/pile/val/part-0-00000.npy
-  #       wikitext_103-validation:
-  #         - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/wikitext_103/val/part-0-00000.npy
+  - label: all-small-ppl-validation
+    data:
+      num_workers: 0
+      drop_last: true
+      # generate_doc_lengths: true
+      memmap_dtype: uint32
+      datasets:
+        c4_en-validation:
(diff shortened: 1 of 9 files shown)


configs/peteish1-weka.yaml
```diff
@@ -108,35 +108,35 @@ eval_interval: 1000
...
+        dolma_wiki-validation:
+          - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/dolma_wiki/val/part-0-00000.npy
+        ice-validation:
+          - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/ice/val/part-0-00000.npy
+        m2d2_s2orc-validation:
+          - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/m2d2_s2orc/val/part-0-00000.npy
+        pile-validation:
+          - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/pile/val/part-0-00000.npy
+        wikitext_103-validation:
+          - /weka/oe-training-default/ai2-llm/eval-data/perplexity/v3_small_dolma2-tokenizer/wikitext_103/val/part-0-00000.npy
```
dirkgr, line 139: Did you mean to commit this change?

configs/peteish1-weka.yaml
```diff
@@ -231,6 +231,228 @@ evaluators:
...
+    type: downstream
+
+  - label: arc_challenge_mc_5shot
+    type: downstream
+
+  - label: arc_challenge_mc_5shot_bpb
+    type: downstream
+
+  - label: copa_rc_0shot
+    type: downstream
```
dirkgr, line 379: Do we care about any of the 0-shots?

olmo/train.py
```diff
@@ -1368,8 +1368,8 @@ def close(self, exit_code: int = 0) -> None:
             gc.enable()
         else:
             gc.disable()
-        if wandb.run is not None:
-            wandb.finish(exit_code=exit_code, quiet=True)
+        # if wandb.run is not None:
+        #     wandb.finish(exit_code=exit_code, quiet=True)
```
dirkgr, line 1372: Debug code?

scripts/eval.py
```diff
@@ -0,0 +1,286 @@
...
+        )
+
+    barrier()
+
+    # Set seed.
+    seed_all(cfg.seed)
+
+    # # Construct data loader.
+    # train_loader = build_train_dataloader(cfg)
+    train_loader = None
```
dirkgr, line 120: Is this always going to be `None`? If so, we don't need it.

scripts/eval.py
```diff
@@ -0,0 +1,286 @@
...
+    # Construct evaluators.
+    evaluators = build_evaluators(cfg, device)
+    barrier()
+
+    if cfg.load_path is None:
+        raise OLMoConfigurationError("To run eval you must provide a load_path")
+    if 'step' in cfg.load_path.split('/')[-1]:
+        load_paths = [cfg.load_path]
+    else:
+        # This globbing does not work with remote paths.
```
dirkgr, line 131: How is that problem handled then?

scripts/eval.py
```diff
@@ -0,0 +1,286 @@
...
+        step = int(load_path.split('/')[-1].split('step')[-1])
+
+        # Initialize the model.
+        log.info("Building model...")
+        olmo_model = OLMo(cfg.model)
+        log.info(f"Total number of parameters: {olmo_model.num_params():,d}")
+        log.info(f"Number of non-embedding parameters: {olmo_model.num_params(include_embedding=False):,d}")
+        log.info(f"Peak GPU Memory (MB) before {cfg.distributed_strategy}: {int(peak_gpu_memory() or 0)}")
+
+        olmo_model.set_activation_checkpointing(cfg.activation_checkpointing)
```
dirkgr, line 144: If we only ever eval, we don't need this.

scripts/eval.py
```diff
@@ -0,0 +1,286 @@
...
+        if param_init_fn is not None or cfg.distributed_strategy == DistributedStrategy.ddp:
+            olmo_model.reset_parameters()
+
+        log.info(f"Peak GPU Memory (MB) after {cfg.distributed_strategy}: {int(peak_gpu_memory() or 0)}")
+        log.info("Model:")
+        log.info(dist_model)
+
+        # Construct optimizer and learning rate scheduler.
+        optim = build_optimizer(cfg, dist_model)
+        scheduler = build_scheduler(cfg)
```
dirkgr, line 226: We don't need optimizers and schedulers if we're just evaluating.

dirkgr: So you're creating these only so that you can produce a `Trainer` object?

How hard is it to pull the stuff you need out of the `Trainer` object, so we don't have to do so many things we don't need? It makes me particularly uncomfortable that you're creating a trainer with a `None` data loader, which isn't supposed to work. It just happens to work.

dirkgr (changes requested):

configs/peteish1-weka.yaml
liujch1998: Yes. I'm under the impression that we should add PPL eval back to peteish training.

configs/peteish1-weka.yaml
liujch1998: Better log them than regret & backfill :)

scripts/eval.py
liujch1998: Not handled. I will assume the checkpoints are on WEKA.

scripts/eval.py
liujch1998: The Trainer class has too many precious helper functions and it's kinda dumb to unroll them all. I do wanna keep at least a dummy Trainer object. Let me see if I can create it w/o the optim/scheduler/etc stuff.

dirkgr: I think you might find that you don't need most of that stuff when you're doing inference only.

scripts/eval.py
dirkgr: At least throw an exception then.

dirkgr:
Let me know when this is ready for another review?

scripts/eval.py
```diff
@@ -0,0 +1,279 @@
...
+    seed_all(cfg.seed)
+
+    # Construct evaluators.
+    evaluators = build_evaluators(cfg, device)
+    barrier()
+
+    if cfg.load_path is None:
+        raise OLMoConfigurationError("To run eval you must provide a load_path")
+    elif "://" in cfg.load_path:
+        raise OLMoConfigurationError("Eval does not support remote paths. Please specify a local path or WEKA mounted path.")
```
liujch1998, line 125: throwing exception for remote paths here

PR #738, opened 2024-10-21
Annealing configs

dirkgr:
This is mainly a bunch of annealing configs that right now live in an obscure branch.
There are also run scripts to make these work in various scenarios.


epwalsh (approved):
That's a lot of configs!

2015aroras (approved):
That's a lot

dirkgr:
It's accumulated over quite a long time. I think we may want to cull these at some point.

PR #739, opened 2024-10-21
Peteish13

dirkgr:
* Peteish13 configs
* More options for `torch.compile()`
* Apply `compile()` to one block at a time.
* Fixes to run in the Google cloud
* Scripts to run on the Augusta cluster


dirkgr:
I think we'll hold off on merging this until Augusta shakes out.

olmo/train.py
```diff
@@ -1036,6 +1036,10 @@ def eval(self) -> Dict[str, Any]:
 
             del eval_batches
 
+        # Eval compiles a bunch more versions, and the result is terrible. This way we get back to zero.
```
epwalsh, line 1039: What do you that the result is terrible?

epwalsh: so this prompted me to look into this a bit more and I think I've found a better solution: just mark the model input sizes as dynamic. I tested this out in OLMo-core and it appears to work well.
https://github.com/allenai/OLMo-core/pull/105

dirkgr: I think it compiles a bunch of versions for different batch sizes, because that's how we call it during eval, and then they stick around. In all of my early runs I had high tps until the first eval, and then low tps afterwards. This is what fixed it.

dirkgr: I tried `dynamic` and it was bad. I don't remember the way in which it was bad, but it didn't work. That's why I added that version in the first place.

epwalsh: Ok, oh well. I tested with nightly so maybe it's just better now with recent compiler advances.

epwalsh (approved):

PR #745, opened 2024-11-12
Documentation Improvements

aman-17:
### Changes Made
- Fixed grammar and improved documentation clarity throughout.
- Restructured training instructions for better readability.
- Enhanced checkpoint download documentation.
  - Added new script `scripts/download_checkpoints.py` to automate checkpoint downloads.
  - Removed manual URL conversion by automating R2 to public URL conversion.
- Fixed bug for loading unsharded checkpoints in `scripts/train.py`.
- Improved data inspection instructions.

### New Features
The new `scripts/download_checkpoints.py` script:
- Automatically handles URL conversions between R2 and public formats.
- Downloads checkpoint files with progress tracking.
- Supports specific step selection and directory listing.


.gitignore
```diff
@@ -31,7 +31,7 @@ pyrightconfig.json
 doc/_build/
 *.swp
 .DS_Store
-
+readme_misc.md
```
dirkgr, line 34: What is this?

README.md
```diff
@@ -100,48 +96,47 @@ python scripts/convert_olmo_to_hf_new.py --input_dir /path/to/olmo/checkpoint --
...
-torchrun --nproc_per_node=8 scripts/train.py configs/official/OLMo-1B.yaml
+python scripts/download_checkpoints.py [PATH_TO_CSV] --save-dir [SAVE_PATH] --step [STEP]
 ```
 
-You can use the same method to launch multi-node jobs as well. See [the documentation](https://pytorch.org/docs/stable/elastic/run.html) for `torchrun` to understand the additional arguments you'll need to configure the rendezvous backend / endpoint.
+Example: To download checkpoint at step 2000:
+```bash
+python scripts/download_checkpoints.py checkpoints/official/OLMo-1B.csv --save-dir ./checkpoints/ --step 2000
+```
+**Note**: All checkpoints in `checkpoints/official/` are unsharded files.
```
dirkgr, line 122:
```suggestion
**Note**: All checkpoints in `checkpoints/official/` are unsharded.
```

dirkgr: They aren't just files. Even in unsharded format, a checkpoint still consists of multiple files.

README.md
```diff
@@ -100,48 +96,47 @@ python scripts/convert_olmo_to_hf_new.py --input_dir /path/to/olmo/checkpoint --
...
 
-To resume training from a checkpoint, you can pass its path (local or URL)
-to `scripts/train.py` with the `--load_path` arguments. For example, to resume training from step 1000 of the OLMo 1B run:
+2. Resume training using the downloaded checkpoint. You can specify either a local path or URL using the --load_path argument: For example, to resume training from step 2000 of the OLMo 1B run:
 
 ```bash
-torchrun --nproc_per_node=8 scripts/train.py configs/official/OLMo-1B.yaml --load_path=https://olmo-checkpoints.org/ai2-llm/olmo-small/w1r5xfzt/step1000-unsharded
+torchrun --nproc_per_node=8 scripts/train.py configs/official/OLMo-1B.yaml     --load_path=checkpoints/step2000 --save_folder=./new_checkpoints --run_name=olmo_test --save_overwrite
```
dirkgr, line 127:
```suggestion
torchrun --nproc_per_node=8 scripts/train.py configs/official/OLMo-1B.yaml --load_path=checkpoints/step2000 --save_folder=./new_checkpoints --run_name=olmo_test
```

README.md
```diff
@@ -100,48 +96,47 @@ python scripts/convert_olmo_to_hf_new.py --input_dir /path/to/olmo/checkpoint --
...
+- Names the training run `olmo_test` in wandb.
+- Overwrites existing checkpoints in the save folder.
 
 ### Inspecting training data
 
-You may be interested in inspecting the exact tokens that composed a particular batch during the training of one of the OLMo models.
-We provide tools to do this, but first you'll need to download the data as above (unless you have an R2 API key) and update the corresponding config accordingly.
-
-Then take note of the URL of the data order file you want, which can be found in the [Models Overview](#models-overview) table. For example, the data order file for the first epoch of the OLMo-7B model is [https://olmo-checkpoints.org/ai2-llm/olmo-medium/wvc30anm/train_data/global_indices.npy](https://olmo-checkpoints.org/ai2-llm/olmo-small/46zc5fly/train_data/global_indices.npy).
+To inspect the exact tokens used in training batches for OLMo models, first download the training data. If you don't have an R2 API key, use the public HTTP URLs and update your configuration file with the local data paths. After completing this setup, you can use the inspection tools to examine the training batches.
```
dirkgr, line 137: Nobody external would ever have an R2 key. I think we can skip that part of the instructions.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,139 @@
...
+    
+    found_files = []
+    for pattern in common_files:
+        test_url = urljoin(url.rstrip('/') + '/', pattern)
+        try:
+            response = requests.head(test_url)
+            if response.status_code == 200:
+                found_files.append(pattern)
+        except requests.exceptions.RequestException:
+            continue
```
dirkgr, line 53: Why would you swallow these exceptions?

dirkgr: Oh, because you don't expect all files to be there? Then at least catch only 404 errors.
But better yet, list the contents of the directory in one call to check what's there, instead of making six calls every time we have to check a directory.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,139 @@
...
+    for file in available_files:
+        file_url = urljoin(public_url.rstrip('/') + '/', file)
+        file_path = base_path / file
+        
+        try:
+            print(f"\nDownloading: {file}")
+            download_file(file_url, file_path)
+        except requests.exceptions.RequestException as e:
+            print(f"Error downloading {file}: {e}")
+            continue
```
dirkgr, line 85: No, wait, don't just catch all exceptions. What about those that can be retried, like timeouts?
Also, there is a risk that this will just skip some files and you don't notice because it prints a tiny message among thousands.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,139 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(description='Download OLMo checkpoints from CSV')
+    parser.add_argument('csv_file', type=str, help='Path to the CSV file containing checkpoint URLs')
+    parser.add_argument('--save-dir', type=str, default='./checkpoints',
+                        help='Base directory to save downloaded checkpoints')
+    parser.add_argument('--step', type=str, help='Specific step number to download (optional)')
+    parser.add_argument('--list-steps', action='store_true', help='List available step numbers and exit')
```
dirkgr, line 93: If you have a tool that can perform multiple different actions, use subcommands.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,139 @@
...
+        public_url = convert_to_public_url(r2_url)
+        print(f"\nStep {step}:")
+        print(f"Original URL: {url}")
+        print(f"R2 URL: {r2_url}")
+        print(f"Public URL: {public_url}")
+    
+    proceed = input("\nDo you want to proceed with the download? (y/n): ")
+    if proceed.lower() != 'y':
+        print("Download cancelled.")
+        return
```
dirkgr, line 129: No, we don't ask for permission. The tools just do the thing. What if we'd want to script it?

However, that means we have to make sure the tools never do anything dangerous by accident.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,139 @@
...
+    
+    for step, url in urls:
+        save_path = os.path.join(args.save_dir, f"step{step}")
+        try:
+            download_checkpoint(url, save_path)
+        except Exception as e:
+            print(f"Error during download of step {step}: {e}")
```
dirkgr, line 136: Do you think anyone will ever want to download all steps? That's a lot of data. I think it's better if we give one command to list steps, and another to download _one_ step, and let them deal with the rest.

scripts/train.py
```diff
@@ -268,9 +268,10 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
                     )
                     cfg.save_num_unsharded_checkpoints_to_keep = cfg.save_num_checkpoints_to_keep
             elif cfg.distributed_strategy == DistributedStrategy.fsdp:
-                checkpoint_type = (
-                    CheckpointType.sharded if cfg.save_num_checkpoints_to_keep != 0 else CheckpointType.unsharded
-                )
+                # checkpoint_type = (
+                #     CheckpointType.sharded if cfg.save_num_checkpoints_to_keep != 0 else CheckpointType.unsharded
+                # )
+                checkpoint_type = CheckpointType.unsharded
```
dirkgr, line 274: What's this?

scripts/train.py
```diff
@@ -297,7 +298,9 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
                 cfg.load_path,
                 load_optimizer_state=not cfg.reset_optimizer_state,
                 load_trainer_state=not cfg.reset_trainer_state,
-                sharded_checkpointer=cfg.load_path_sharded_checkpointer,
+                # sharded_checkpointer=cfg.load_path_sharded_checkpointer,
+                sharded_checkpointer= False,
+                checkpoint_type=CheckpointType.unsharded
```
dirkgr, line 303: Same question here

dirkgr (changes requested):

.gitignore
aman-17: This was my read_me with grammatical mistakes. I will remove it from .gitignore.

README.md
aman-17: Without --save_overwrite, the program throws error.

dirkgr: Only if the directory already exists

scripts/train.py
aman-17: I have added this because in `if cfg.load_path is not None:` we are passing the unsharded checkpoints which we download from r2. But `sharded_checkpointer=cfg.load_path_sharded_checkpointer,` this is taking sharded instead of unsharded. I had to manually set the flag to `False`.

scripts/download_checkpoints.py
aman-17: For each step we have only 4 files(not a lot). You want me to add timeout exception and retry it again?

scripts/download_checkpoints.py
aman-17: The website does not support directory listing, so I had to manually add these ("config.yaml", "model.pt", "optim.pt", "train.pt", "model.safetensors", "optim.safetensors")

scripts/download_checkpoints.py
aman-17: Added timeout exception with the help of GPT.

scripts/download_checkpoints.py
aman-17: Removed this

aman-17 (commented):
@dirkgr Can you review this?

README.md
aman-17: Updated

README.md
aman-17: Updated this

README.md
```diff
@@ -58,7 +55,7 @@ The core models in the OLMo family released so far are (all trained on the [Dolm
 URLs to checkpoints at intermediate steps of the models' trainings can be found in the csv files under [`checkpoints/official/`](https://github.com/allenai/OLMo/blob/main/checkpoints/official). These 'directory' URLs cannot currently be directly accessed, but files within the directory are publicly accessible. These URLs can also be provided to the training script to resume training from the checkpoint (see [Training](#training)). Each checkpoint directory consists of:
 
 - `config.yaml`: the config at that training step.
-- `model.pt`, `optim.pt`, `train.pt`: model, optimizer and training state at that training step.
+- `model.safetensors`, `optim.safetensors`, `train.pt`: model, optimizer and training state at that training step.
```
2015aroras, line 58: `train.safetensors`? Also, for the original model we just have `*.pt` so we should have that format mentioned somewhere.

scripts/download_checkpoints.py
2015aroras: I think Dirk is referring to this: https://docs.python.org/3/library/argparse.html#sub-commands .

README.md
aman-17: We are going to save in .safetensors starting from OLMo-2

scripts/download_checkpoints.py
dirkgr: Ok, manually looking for them works too.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+        try:
+            test_url = urljoin(url.rstrip('/') + '/', pattern)
+            response = requests.head(test_url)
+            # response.raise_for_status()
+            if response.status_code == 200:
+                found_files.append(pattern)
+        except requests.exceptions.HTTPError as e:
+            print(f"HTTP error for {pattern}: {e}")
+        except requests.exceptions.RequestException as e:
+            print(f"Connection error for {pattern}: {e}")
```
dirkgr, line 55: This is still swallowing exceptions. It just prints about them. This part of the code should check whether the exception is a 404 error, in which case it's fine, and otherwise let the exception propagate.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+def download_checkpoint(url, save_dir):
+   """Download all files from a checkpoint directory."""
+   r2_url = convert_to_r2_url(url)
+   public_url = convert_to_public_url(r2_url)
+   base_path = Path(save_dir)
+   base_path.mkdir(parents=True, exist_ok=True)
+   print(f"Saving to: {base_path}")
+   available_files = try_get_directory_listing(public_url)
+   
+   if not available_files:
```
dirkgr, line 67:
```suggestion
   if len(available_files) <= 0:
```

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+        "train.pt",
+        "model.safetensors",
+        "optim.safetensors",
+    ]
+    found_files = []
+    for pattern in common_files:
+        try:
+            test_url = urljoin(url.rstrip('/') + '/', pattern)
+            response = requests.head(test_url)
+            # response.raise_for_status()
```
dirkgr, line 49: Leftover debugging code?

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+       except requests.exceptions.Timeout:
+           print(f"Timeout error for {file}, retrying once...")
+           try:
+               download_file(file_url, file_path)
+           except requests.exceptions.RequestException as e:
+               failed_files.append(file)
+               print(f"Failed to download {file}: {e}")
+       except requests.exceptions.RequestException as e:
+           failed_files.append(file)
+           print(f"Failed to download {file}: {e}")
```
dirkgr, line 86: Don't retry in this way. The Python requests library already supports retries. In this code we should just call `download_file()`. If a file fails, we fail the whole process (i.e., we let the exception propagate).

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+    response.raise_for_status()
+    total_size = int(response.headers.get('content-length', 0))
+    save_path.parent.mkdir(parents=True, exist_ok=True)
+    
+    with open(save_path, 'wb') as f:
+        with tqdm(total=total_size, unit='B', unit_scale=True, desc=save_path.name) as pbar:
+            for chunk in response.iter_content(chunk_size=chunk_size):
+                if chunk:
+                    f.write(chunk)
+                    pbar.update(len(chunk))
```
dirkgr, line 33: Retries should be handled in here. If the request fails with a recoverable error code ([anything that starts with 5XX, 408, 409, 429](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)), it should wait one second, and then try again. Try up to 5 times, and then give up. When giving up, it must make sure that the file at `save_path` does not exist.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+                               help='Base directory to save downloaded checkpoints')
+    list_parser = subparsers.add_parser('list',
+                                       help='List available checkpoint steps')
+    list_parser.add_argument('csv_file', type=str,
+                            help='Path to the CSV file containing checkpoint URLs')
+    args = parser.parse_args()
+    
+    print(f"Reading CSV file: {args.csv_file}")
+    
+    with open(args.csv_file, 'r') as f:
```
dirkgr, line 109: That looks like `csv_file` is a required argument.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+        for step, _ in urls:
+            print(f"Step {step}")
+        return
+    
+    if args.step:
+        urls = [(step, url) for step, url in urls if step == args.step]
+        if not urls:
+            print(f"Error: Step {args.step} not found in the CSV file.")
+            print("Use list argument to see available step numbers.")
+            return
```
dirkgr, line 124: Don't print and then return. If you do this, the process will return normally, and whoever called it will think it succeeded. If you want to fail, raise an exception and let it crash the process.

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+                            help='Path to the CSV file containing checkpoint URLs')
+    args = parser.parse_args()
+    
+    print(f"Reading CSV file: {args.csv_file}")
+    
+    with open(args.csv_file, 'r') as f:
+        reader = csv.DictReader(f)
+        urls = [(row['Step'], row['Checkpoint Directory']) for row in reader]
+    
+    if args.command == 'list':
```
dirkgr, line 113: This should be an if ladder:
```
if args.command == 'foo':
    do_foo_things()
    do_other_things()
elif args.command == 'bar':
    do_bar_things()
    do_more_things()
else:
    raise RuntimeException(f"Unexpected command: {args.command}")
```

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+        print("Available steps:")
+        for step, _ in urls:
+            print(f"Step {step}")
+        return
+    
+    if args.step:
```
dirkgr, line 119:
```suggestion
    if args.step is not None:
```

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+       print(f"\nWARNING: Failed to download these files: {failed_files}")
+
+def main():
+    parser = argparse.ArgumentParser(description='Download OLMo checkpoints')
+    subparsers = parser.add_subparsers(dest='command', help='Commands')
+    download_parser = subparsers.add_parser('download', 
+                                          help='Download checkpoints from CSV file')
+    download_parser.add_argument('csv_file', type=str, 
+                               help='Path to the CSV file containing checkpoint URLs')
+    download_parser.add_argument('--step', type=str, required=True,
```
dirkgr, line 97: In the code it seems like `step` is not required if you want to download everything?

scripts/download_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+        urls = [(step, url) for step, url in urls if step == args.step]
+        if not urls:
+            print(f"Error: Step {args.step} not found in the CSV file.")
+            print("Use list argument to see available step numbers.")
+            return
+    
+    print(f"Saving checkpoints to: {args.save_dir}")
+    for step, url in urls:
+        r2_url = convert_to_r2_url(url)
+        public_url = convert_to_public_url(r2_url)
```
dirkgr, line 129: The URLs in the CSV are already public URLs? Why are we doing this change?

dirkgr (changes requested):

scripts/download_checkpoints.py
aman-17: How does this make any change? found_files can never be negative right? `if not available_files:` is also same as `if len(available_files) == 0:`.

scripts/download_checkpoints.py
aman-17: No, step is required argument. It throws this error when user fails to add `--step`.
```bash
python scripts/download_checkpoints.py download checkpoints/official/OLMo-1B.csv  --save-dir ./new_checkpoints 
usage: download_checkpoints.py download [-h] --step STEP [--save-dir SAVE_DIR] csv_file
download_checkpoints.py download: error: the following arguments are required: --step
```

scripts/download_checkpoints.py
aman-17: I missed adding `required=True` for csv_file.

aman-17: Yes, without csv we can't download the checkpoints since it contains links.

scripts/download_checkpoints.py
aman-17: I've taken these functions from my another file where I will have r2 urls in csv while uploading. I missed this here.

README.md
2015aroras: Sure, but people might still try to use older OLMo models. The documentation should be backwards-compatible?

dirkgr: No, not backwards compatible

PR #747, opened 2024-11-18
Add intermediate size to hf_olmo

2015aroras:
`intermediate_size` is a commonly used field in transformers. We should support it in `hf_olmo`. The current vLLM implementation for `hf_olmo` does a try-except to handle the absence of this field.


2015aroras:
Not tested but this is a pretty straightforward change and I'll fix anything that breaks

2015aroras:
Closing as I will probably need to add a few more things.

PR #748, opened 2024-11-18
Add test and train sets to in-loop oe-eval (for ladder work)

liujch1998:
Standardizing what we should eval for the ladder work:
* 10 tasks from OLMES
* val sets, and (when applicable) test and/or train sets
* No subsampling happens
* rc_5shot, mc_5shot, and their bpb version (for MMLU, I kept a rc_var version)

Stats:
Screenshot 2024-11-18 at 15 23 03


olmo/eval/downstream.py
```diff
@@ -1863,3 +1863,503 @@ def doc_to_label(self, doc) -> int:
...
+        {"dataset_path": "arc_challenge", "dataset_name": "test_mc_5shot", "metric_type": "acc"},
+    ),
+    "arc_challenge_test_mc_5shot_bpb": (
+        OEEvalTask,
+        {"dataset_path": "arc_challenge", "dataset_name": "test_mc_5shot", "metric_type": "bpb"},
+    ),
+    "arc_easy_train_rc_5shot": (
+        OEEvalTask,
+        {"dataset_path": "arc_easy", "dataset_name": "train_rc_5shot", "metric_type": "len_norm"},
+    ),  # this used to be acc
```
liujch1998, line 1922: Can we safely change acc to len_norm here? I want to match what arc_challenge is doing.

OyvindTafjord (approved):
LGTM, I tried to look through the tasks, didn't spot anything suspicious. (The number of tasks is getting pretty unwieldy at this point, would be good to streamline in the future somehow)

PR #752, opened 2024-11-26
Add OLMo 2 checkpoint converter and update docs

2015aroras:


docs/Checkpoints.md
```diff
@@ -21,7 +21,7 @@ Unless otherwise specified, an OLMo checkpoint is assumed to be unsharded. OLMo
 Each unsharded checkpoint directory consists of:
 
 - `config.yaml`: the config at that training step.
-- `model.pt`, `optim.pt`, `train.safetensors`: model, optimizer and training state at that training step.
+- `model.safetensors`, `optim.safetensors`, `train.safetensors`: model, optimizer and training state at that training step. Checkpoints of older OLMo releases use the `.pt` extension instead.
```
dirkgr, line 24: I actually think `train` still uses .pt no matter what. The only one I updated was wrong :-(
```suggestion
- `model.safetensors`, `optim.safetensors`, `train.pt`: model, optimizer and training state at that training step. Checkpoints of older OLMo releases use the `.pt` extension instead.
```

dirkgr (approved):
Approve after the one fix

PR #768, opened 2024-12-18
Update README.md

codeviking:
I think formatting the path as code makes things a bit more presentable. This is a very small change :).


cnewell (commented):
Looks fine to me. Deferring to Pete and the OLMo team on the decision to merge or not.

epwalsh (approved):

PR #769, opened 2024-12-21
enable non-distributed training and MPS support

peter-sk:
Current OLMo is locked to DDP or FSDP distributed training on CUDA accelerators.

This PR does two things:
1) It adds a single accelerator mode (distributed_strategy: single) through a simple wrapper module.
2) It adds support for MPS, allowing development and debugging on ARM64 Mac machines.

The primary motivation is to make and test code changes without having to block GPU resources. In addition, tiny models can actually be trained on a sufficiently equipped system.


dirkgr:
We just merged a version of MPS training that works a little different than yours, but should also solve the problem. Thanks for this one though, we took inspiration from it!

PR #771, opened 2024-12-24
Decoupled Momentum Optimization

peter-sk:
Cleaned-up version of https://github.com/bloc97/DeMo for integrating efficient distributed training a la Decoupled Monentum Optimization (https://arxiv.org/abs/2411.19870)

### DIFF
--- a/olmo/config.py
+++ b/olmo/config.py
@@ -498,6 +498,7 @@ class ModelConfig(BaseConfig):
 class OptimizerType(StrEnum):
     lionw = "lionw"
     adamw = "adamw"
+    demo = "demo"
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


olmo/optim.py
```diff
@@ -961,6 +1134,17 @@ def build_optimizer(cfg: TrainConfig, model: nn.Module) -> Optimizer:
...
             eps=cfg.optimizer.eps,
         )
+    elif cfg.optimizer.name == OptimizerType.demo:
+        return DeMo(
+            param_groups,
+            compression_decay=cfg.optimizer.compression_decay,
+            compression_topk=cfg.optimizer.compression_topk,
+            compression_chunk=cfg.optimizer.compression_chunk,
+            weight_decay=cfg.optimizer.weight_decay,
+            process_group=None,  # TODO: fix for hybrid sharding
```
dirkgr, line 1144: This seems important? Hybrid is necessary for big models.

olmo/config.py
```diff
@@ -830,6 +851,8 @@ class FSDPConfig(BaseConfig):
     PyTorch's default HSDP behavior matches this default behavior.
     """
 
+    disable_grad_sync: bool = False
+
```
dirkgr, line 855: I see this setting twice, once here, and once in `DDPGradSyncMode`?

olmo/optim.py
```diff
@@ -647,6 +649,177 @@ def get_post_step_metrics(
             return metrics
 
 
+class DeMo(torch.optim.SGD, Optimizer):
```
dirkgr, line 652: It seems like the organization would make more sense if this class, and `demo_utils.py`, were in their own file together, and then we use `__all__` to make this optimizer appear the same as the others.

dirkgr (changes requested):
This looks interesting, but I don't actually know what this optimizer is. Can you give some background?

dirkgr:
Oh, I see. You put a reference in the description 🙈.

Paper says you pushed this to 1B/100B tokens. Can you go further? Experience says, things like this stop working if you go really big.

Issue #780, opened 2025-01-11
Allow arbitrary implementations of pytorch datasets to be used/specified in the configuration file


mbsabath:
### 🚀 The feature, motivation and pitch

I'm currently working within the Kempner Institute at Harvard at preparing an internal dataset management system (and associated library) that will allow researchers to find and use data with multiple LLM frameworks on an on prem HPC system with a single API. As part of this package, we provide a way to create a Pytorch Dataset object for our data that can be used for training. We would like to handle our own implementation of the dataset to avoid external dependencies on multiple frameworks/to allow researchers space to develop their own frameworks/to avoid having to make slightly different copies of the same data for all frameworks researchers want to use.

Olmo as currently written only supports `IterableDatasets` as the input format, which only currently works with the `MemmapDataset` objects that can be specified as lists of file paths. I propose adding a configuration option to allow users to optionally specify a custom dataset class and its arguments. I also propose including additional configuration options allowing users to specify how to map from the outputs of that dataset to the expected fields for Olmo's collation function (or to provide their own collation function).

I have a currently written PR implementing this functionality. I've tested it running a 7b model using FSDP and confirmed that current OLMo configurations will run successfully without changes, while we successfully training proceed as expected when using the custom dataset. I've added appropriate logging statements to make clear that custom datasets may not support the deterministic training that OLMo's `IterableDataset` offers.

### Alternatives

An alternative approach could instead be to add functionality to support wrapping arbitrary dataset objects in an `IterableDataset` object but I believe that supporting that would add unneeded complexity to the `IterableDataset` class rather than letting external users/developers handle the complexity of adapting their data to OLMo.

aman-17:
Hey @mbsabath, thank you for your contributions. we’ll review your PR and test it to ensure compatibility with existing configurations and workflows. If everything checks out, we’ll proceed with merging it. The added logging for transparency about deterministic training is also a great touch :)

PR #781, opened 2025-01-11
Add configuration option to allow users to specify a custom Dataset class

mbsabath:
solution to #780 

Adds an optional field to the data configuration to allow users to specify an arbitrary class to use 
as their underlying dataset to be wrapped in the OLMo `IterableDataset`. Allows users to also specify their own collation function (assuming that it will output fields that OLMo proper can work with) or to specify mappings from fields in their dataset to fields expected by OLMo's collation function. 

Unit tests have been added, and integration tests have been run confirming that no currently existing OLMo configurations will need to be changed and that both standard OLMo dataset based training runs and custom dataset runs are able to run to completion and be restarted successfully from checkpoints. For example use, please see the included example config in this PR.

Edit: Updated PR to wrap custom dataset in an `IterableDataset`, updated the PR description to reflect this


olmo/data/collator.py
```diff
@@ -138,3 +138,37 @@ def __call__(self, items: Union[List[Dict[str, Any]], List[torch.Tensor]]) -> Di
...
+
+    def _relabel_item(self, item: Dict[str, Any]) -> Dict[str, Any]:
+        out = {
+            "input_ids": item.__getitem__(self.input_id_field),
+            "attention_mask": item.__getitem__(self.attention_mask_field) if self.attention_mask_field else None,
+            "attention_bias": item.__getitem__(self.attention_bias_field) if self.attention_bias_field else None,
+            "label_mask": item.__getitem__(self.label_mask_field) if self.label_mask_field else None,
+            "index": item.__getitem__(self.index_field) if self.index_field else None,
+            "instance_mask": item.__getitem__(self.instance_mask_field) if self.instance_mask_field else None,
+            "metadata": item.__getitem__(self.metadata_field) if self.metadata_field else None,
```
dirkgr, line 165: Why all these `item.__getitem__` calls? Why not just `item[]`?

dirkgr (commented):
By and large fine with, this just one question.

olmo/data/collator.py
mbsabath: I can change this, was intially working trying to mirror the `.get` method used in the default collator, agree that `[]` is cleaner, can make the change

mbsabath:
Gonna fix the changelog issue today and address @dirkgr's comments, anything in particular I should do to address the failing actions or should merging in changes from the main branch address the issue?

olmo/data/collator.py
mbsabath: Fixed

dirkgr (approved):
Looks good to me. Let's see if the checks pass this time and then we're good to go.

PR #789, opened 2025-01-22
MPS support

aman-17:
Added MPS support.


dirkgr:
Is this all good to go then?

aman-17:
Yeah, just removed precision and batch_size from main function as you've mentioned last week in the meeting.

olmo/checkpoint.py
```diff
@@ -645,7 +646,7 @@ def save_checkpoint(
                     self._write_optim_dict(
                         optim_state_dict, checkpoint_dir, upload_to, save_overwrite=self.cfg.save_overwrite
                     )
-            elif isinstance(dist_model, DDP):
+            elif isinstance(dist_model, DDP) or isinstance(dist_model, SingleAccelerator):
```
dirkgr, line 649:
```suggestion
            elif isinstance(dist_model, (DDP, SingleAccelerator)):
```

dirkgr: You can do this in every place where you check for multiple different classes.

olmo/config.py
```diff
@@ -831,6 +837,31 @@ class FSDPConfig(BaseConfig):
...
+            elif torch.cuda.is_available():
+                return torch.device("cuda")
+            else:
+                return torch.device("cpu")
+        elif self.device == "mps" and not torch.backends.mps.is_available():
+            print("MPS not available.")
+            return torch.device("cpu")
+        elif self.device == "cuda" and not torch.cuda.is_available():
+            print("CUDA not available.")
+            return torch.device("cpu")
```
dirkgr, line 860: If the device is explicitly set to something that won't work, we should just fail instead of silently fixing it. Throw an exception in this case, like `OLMoConfigurationError`.

We should _never_ just print something. Sometimes we log something (through the logging mechanism). If we need things to fail, we throw exceptions. Also, printing something will go to stdout, where it does not belong in the case of failures.

olmo/model.py
```diff
@@ -519,6 +523,8 @@ def _cast_attn_bias(cls, bias: torch.Tensor, input_dtype: torch.dtype) -> torch.
             target_dtype = torch.get_autocast_gpu_dtype()
         elif bias.device.type == "cpu" and torch.is_autocast_cpu_enabled():
             target_dtype = torch.get_autocast_cpu_dtype()
+        elif bias.device.type == "mps":
+            target_dtype = torch.float32
```
dirkgr, line 527: No. My mps device, for example, supports bfloat16. So we should not just silently overwrite things.

olmo/model.py
```diff
@@ -1747,7 +1753,7 @@ def from_checkpoint(
         if checkpoint_type == CheckpointType.unsharded:
             # Initialize model (always on CPU to start with so we don't run out of GPU memory).
             model_config.init_device = "cpu"
-            model = OLMo(model_config)
+            model = OLMo(model_config).to(device)
```
dirkgr, line 1756: Did we talk about this? I'm pretty sure that doing this `.to(device)` thing will OOM for large distributed models.

olmo/torch_util.py
```diff
@@ -75,7 +75,7 @@ def move_to_device(o: T, device: torch.device) -> T:
     elif isinstance(o, tuple):
         return tuple((move_to_device(x, device) for x in o))  # type: ignore[return-value]
     else:
-        return o
+        return o.to(device)  # type: ignore
```
dirkgr, line 78: Why did you do this? What case is this handling? This final `else` is supposed to handle the case of other objects that don't have a `.to()` method.

olmo/train.py
```diff
@@ -430,7 +432,10 @@ def restore_rng_state(self, rng_state: Dict[str, Any]) -> None:
         random.setstate(rng_state["python"])
         np.random.set_state(rng_state["numpy"])
         torch.set_rng_state(rng_state["torch"])
-        torch.cuda.set_rng_state(rng_state["cuda"])
+        if rng_state["cuda"] is not None:
+            torch.cuda.set_rng_state(rng_state["cuda"])
+        if rng_state["mps"] is not None:
+            torch.mps.set_rng_state(rng_state["mps"])
```
dirkgr, line 438: This needs to check whether those devices are available. We might be taking a model that was trained on a GPU and try to continue training it on a MPS.

dirkgr: If a device is available, but we don't have any random state to set for it, we should print a warning.

olmo/train.py
```diff
@@ -796,7 +801,7 @@ def train_batch(self, batch: Dict[str, Any]) -> Tuple[torch.Tensor, Optional[tor
             output_hooks += self._setup_module_output_save_hooks(micro_batch_idx)
 
             with grad_sync_context():
-                with torch.autocast("cuda", enabled=True, dtype=self.cfg.autocast_precision):
+                with torch.autocast("mps", enabled=True, dtype=self.cfg.autocast_precision):
```
dirkgr, line 804: This is going to break all training on GPU?

scripts/train.py
```diff
@@ -130,7 +137,7 @@ def main(cfg: TrainConfig) -> None:
 
     # Initialize the model.
     log.info("Building model...")
-    olmo_model = OLMo(cfg.model)
+    olmo_model = OLMo(cfg.model).to(device)
```
dirkgr, line 140: Same with the `to()` here. I don't think we can do that at this point.

scripts/train.py
```diff
@@ -211,8 +218,10 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
             param_init_fn=param_init_fn,
             **hybrid_sharding_fsdp_kwargs,
         )
-    elif cfg.distributed_strategy is None:
-        raise NotImplementedError("Single accelerator training not implemented yet!")
+    elif cfg.distributed_strategy == DistributedStrategy.single:
+        param_init_fn = None
+        olmo_model = olmo_model.to(device)
+        dist_model = SingleAccelerator(olmo_model)
```
dirkgr, line 224: We still need to check for `None` somehow. If it's `None`, we throw a config exception.

scripts/train.py
```diff
@@ -387,4 +424,8 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
         raise OLMoCliError(f"Usage: {sys.argv[0]} [CONFIG_PATH] [OPTIONS]")
 
     cfg = TrainConfig.load(yaml_path, [clean_opt(s) for s in args_list])
+    if mps_device:
+        log.info("Device is MPS. Updating config...")
+        cfg.model.init_device = "mps"
+        cfg.distributed_strategy = "single"  # type: ignore
```
dirkgr, line 430: We also need these kinds of overrides for the case of the CPU. I think it will be easier if instead of having a flag like `mps_device`, you check what's in the torch default device.

dirkgr (changes requested):

olmo/model.py
aman-17: Hey Dirk, I explicitly assigned float32 because `torch.get_autocast_mps_dtype()` is not available in Pytorch and autocast on MPS defaults to float16. I chose float32 to avoid gradient vanishing issues during backprop.
Link: https://github.com/pytorch/pytorch/issues/88415

olmo/train.py
```diff
@@ -432,10 +432,16 @@ def restore_rng_state(self, rng_state: Dict[str, Any]) -> None:
...
         torch.set_rng_state(rng_state["torch"])
-        if rng_state["cuda"] is not None:
-            torch.cuda.set_rng_state(rng_state["cuda"])
-        if rng_state["mps"] is not None:
-            torch.mps.set_rng_state(rng_state["mps"])
+        if torch.cuda.is_available():
+            if rng_state["cuda"] is not None:
+                torch.cuda.set_rng_state(rng_state["cuda"])
+            else:
+                print("Warning: CUDA is available, but no RNG state was provided.")
```
dirkgr, line 439: Don't print! Use the logger instead.

scripts/train.py
```diff
@@ -424,8 +424,12 @@ def dummy_init_fn(module: torch.nn.Module) -> None:
...
     cfg = TrainConfig.load(yaml_path, [clean_opt(s) for s in args_list])
-    if mps_device:
+    if torch.device("mps"):
         log.info("Device is MPS. Updating config...")
         cfg.model.init_device = "mps"
         cfg.distributed_strategy = "single"  # type: ignore
+
+    if torch.device("cpu"):
+        log.info("Device is CPU. Updating config...")
+        cfg.model.init_device = "cpu"
```
dirkgr, line 434: This branch also needs the override for `distributed_strategy`.

dirkgr (changes requested):
Did you try this to make sure it still works on MPS, and also on GPS?

olmo/train.py
```diff
@@ -436,12 +436,12 @@ def restore_rng_state(self, rng_state: Dict[str, Any]) -> None:
             if rng_state["cuda"] is not None:
                 torch.cuda.set_rng_state(rng_state["cuda"])
             else:
-                print("Warning: CUDA is available, but no RNG state was provided.")
+                log.info("Warning: CUDA is available, but no RNG state was provided.")
```
dirkgr, line 439: Should be `log.warning`, since these are warnings.

dirkgr (changes requested):

aman-17:
> Did you try this to make sure it still works on MPS, and also on GPS?

It runs fine on both MPS and GPU.
Our `.toml` file installs Pytorch 2.6.0, but we need to downgrade to 2.5.1 for training.

dirkgr:
> Our .toml file installs Pytorch 2.6.0, but we need to downgrade to 2.5.1 for training.
Why is that?

olmo/model.py
dirkgr: This was a rabbit hole, but you can just use `torch.get_autocast_dtype("mps")`.

dirkgr (changes requested):

dirkgr:
> we changed the default value of the `weights_only`

Can you just change it where we call the function? Pass in the old value?

aman-17:
> > we changed the default value of the `weights_only`
> 
> Can you just change it where we call the function? Pass in the old value?

Passed `weights_only=False` explicitly. Works well with both the torch versions(2.5.x and 2.6.0)

olmo/model.py
```diff
@@ -1762,7 +1768,7 @@ def from_checkpoint(
                 )
 
                 model_config.init_device = device
-                model = OLMo(model_config)
+                model = OLMo(model_config).to(device)
```
dirkgr, line 1771: We still have these lines, but I'm worried this won't work when the model is on the "meta" device, and isn't supposed to be initialized until later. Can you try that scenario? Try training a 7B model on 8 GPUs, just for a step or two to see if it still works?

dirkgr (changes requested):
One more thing

olmo/model.py
aman-17: Works well on 8 GPU, I tested on A6000 as well (decreased the batch size by half to avoid OOM). However, I removed these `.to(device)` just to be sure the model isn’t placed on device before checkpoint loading.

dirkgr (approved):

dirkgr:
Thank you!

PR #795, opened 2025-02-12
Updated default value of memmap_dtype to uint32

aman-17:
Updated `memmap_dtype` to uint32 for compatibility with OLMo-2-1124.
Added links to the Discord and Playground in the readme.
Minor improvements in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,12 @@
   <a href="https://arxiv.org/pdf/2501.00656.pdf">
     <img alt="Paper URL" src="https://img.shields.io/badge/arxiv-2402.00838-blue">
   </a>
+  <a href="https://playground.allenai.org">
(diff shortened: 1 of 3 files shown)


dirkgr (changes requested):
Sorry, I think you misunderstood. What we want is, set the dtype based on the vocab size. If the vocab size is > `2**16`, then the dtype has to be `uint32`. If it's smaller, it has to be `uint16`.

dirkgr (changes requested):
The changelog is no longer accurate.

But my main question is, what happens when someone overrides the setting? Is that possible? It should be.

There is still `data_config.effective_memmap_dtype`. Where does that come from? Do we have to respect if it's set?

Issue #798, opened 2025-02-18
Failed to resolve dependency while using uv


aztecher:
### 🐛 Describe the bug

Hi there!
I have found that when I try to setup using [uv](https://github.com/astral-sh/uv), it fails to resolve the dependencies.
The error is as follows

```
Using CPython 3.12.9
Creating virtual environment at: .venv
  × No solution found when resolving dependencies for split (python_full_version == '3.8'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy Python>3.8 and
      cached-path>=1.6.2 depends on Python>3.8, we can conclude that cached-path>=1.6.2
      cannot be used.
      And because only the following versions of cached-path are available:
          cached-path<=1.6.2
          cached-path==1.6.3
          cached-path==1.6.4
          cached-path==1.6.5
          cached-path==1.6.6
          cached-path==1.6.7
      we can conclude that cached-path>=1.6.2 cannot be used.
      And because your project depends on cached-path>=1.6.2 and your project requires
      ai2-olmo[all], we can conclude that your project's requirements are unsatisfiable.

      hint: The `requires-python` value (>=3.8) includes Python versions that are not
      supported by your dependencies (e.g., cached-path>=1.6.2 only supports >3.8). Consider
      using a more restrictive `requires-python` value (like >3.8).
```

This error occurs because the OLMo dependency, cached-path, has [requires-python = ">3.8"](https://github.com/allenai/cached_path/blob/v1.6.2/pyproject.toml#L20), while OLMo itself has [requires-python = ">=3.8"](https://github.com/allenai/OLMo/blob/main/pyproject.toml#L13).
Therefore, changing OLMo's `requires-python` to `>3.8` will solve the problem (I have confirmed that this change solves the problem in my environment).

If you don't have a strong reason to set `requires-python = ">=3.8"` in this repository, It would be better to change `requires-python = ">3.8"`.
What do you think of this change? If ok, I will create PR to fix this issue.

Thanks ;)

**How to reproduce**

- python: 3.12
- uv: 0.6.0

```bash
uv python install 3.12
uv python pin 3.12
git clone https://github.com/allenai/OLMo.git
cd OLMo
uv sync --extra all
```

### Versions

```
$ python --version && pip freeze
Python 3.12.0

```

NOTE: [uv](https://github.com/astral-sh/uv) creates a new environment with the specified version of Python. So, a specified Python version is displayed and this environment has no packages.

aman-17:
[...]

PR #800, opened 2025-02-22
Fix the dependency problem

aztecher:
Fix the dependency problem reported in https://github.com/allenai/OLMo/issues/798


dirkgr:
Looks like somebody else had already made this change. I'll still try to merge this for credit assignment :-)

dirkgr (approved):

PR #803, opened 2025-02-26
Update olmo_core version

AkshitaB:
- Update `olmo_core` version
- Remove python 3.8 support


.github/workflows/main.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.8', '3.10']
+        python: ['3.10']
```
2015aroras, line 31: thought: add 3.12? Can't hurt right?

2015aroras (approved):

2015aroras (approved):

.github/workflows/main.yml
AkshitaB: There are no distributions for torch2.2 with python3.12. This can be dealt with in a separate PR when we're feeling up to upgrading the torch version (will need to update docker image, etc).

PR #804, opened 2025-02-28
Fix device detection logic in train.py

MohdElgaar:
This PR fixes a bug in the device detection logic in train.py.

## Issue
The code incorrectly used `torch.device()` to check for device availability, which always evaluates to `True` in a boolean context. This caused both MPS and CPU conditions to always be true, resulting in the configuration always ending with `cfg.model.init_device = "cpu"` and `cfg.distributed_strategy = "single"`.

## Fix
Changed to use proper availability checks with `torch.backends.mps.is_available()` and `torch.cuda.is_available()`.

## Note
Despite this bug, training has worked because the actual device allocation happens correctly in the `main()` function

The fix ensures correct configuration from the start, avoiding initialization in-efficiency.


2015aroras (approved):
Thank you!

PR #811, opened 2025-03-21
[Feature] add tensorbard support to the training framework

lyuwen:
Hi, I was using the olmo framework to learn pretraining on LLM and I found that it only support wandb as output for training metrics. In my training environment, the connection to wandb is not as stable, and I switched to tensorboard for local logging. It works quite well. Thus I am creating this pull request to share this modification.

### DIFF
--- a/olmo/config.py
+++ b/olmo/config.py
@@ -1200,6 +1200,11 @@ class TrainConfig(BaseConfig):
     Weights & Biases configuration.
     """
 
+    log_to_tensorbard: Optional[str] = None
+    """
+    Path to tensorbard log output directory.
+    """
+
     speed_monitor: SpeedMonitorConfig = field(default_factory=SpeedMonitorConfig)
     """
     Speed monitor configuration.
--- a/olmo/tensorbard_logger.py
+++ b/olmo/tensorbard_logger.py
@@ -1,0 +1,56 @@
+import os.path as osp
+from torch.utils.tensorboard import SummaryWriter
+
+
+# create a new class inheriting from SummaryWriter
+class NewSummaryWriter(SummaryWriter):
+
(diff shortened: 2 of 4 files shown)


olmo/config.py
```diff
@@ -1200,6 +1200,11 @@ class TrainConfig(BaseConfig):
     Weights & Biases configuration.
     """
 
+    log_to_tensorbard: Optional[str] = None
```
dirkgr, line 1203: Since this takes a path, it should be called "tensorboard_path" or something like that.

olmo/tensorbard_logger.py
```diff
@@ -0,0 +1,56 @@
...
+    # and uses built-in add_scalar() function
+    # that function combines all plots into one subgroup by a tag
+    def add_scalar_dict(self, dictionary, global_step, tag=None):
+        for name, val in dictionary.items():
+            if tag is not None:
+                name = osp.join(tag, name)
+            self.add_scalar(name, val, global_step)
+
+
+writer = None
```
dirkgr, line 22: Can you restructure this so that you don't need a global object here, and instead we have the `writer` object in `train.py`, which can be `None` if tensorboard is disabled, and is an object otherwise?

dirkgr (changes requested):
How does this deal with the additional dependency? What happens when you don't have tensorboard installed?

olmo/tensorbard_logger.py
lyuwen: Thank you for you advice, I just pushed an update that makes the logger an attribute of the Trainer class and handled the situation where dependency is not found.

olmo/config.py
lyuwen: Sorry I missed this at first. It is now fixed.

PR #821, opened 2025-04-07
Fix tokenizer.encode() to respect add_special_tokens=False parameter

### LINKED ISSUE #765: tokenizer.encode function`s param add_special_tokens=False not work.
### 🐛 Describe the bug

the tokenizer is from olmo.tokenizer package.
Keep token id to the default value 50279,  when the default tokenizer is loaded, run the code below:
input：
`tokenizer.encode("hello", add_special_tokens=False) ` 
output :  
[25521, 50279]
The Result shows that the parameter 'add_special_tokens=False' dose not work.
And I find the reason is in /olmo/tokenizer.py  line 183:
`batch_encoding = self.base_tokenizer.encode_batch(inputs)`
the param 'add_special_tokens' didn't pass to the base_tokenizer's encode function.
tokenizer_error

I can find the bug because it caused an assertion in /scripts/prepare_tulu_data.py line 90
prepare_tulu_l90

### Versions

0.5.1

ved1beta:
## Problem
The `add_special_tokens=False` parameter in the tokenizer's `encode`/`encode_batch` methods doesn't work as expected. Even when set to `False`, special tokens (like EOS) are still being added to the encoded output.

## Root Cause
The issue occurs because the `add_special_tokens` parameter is not being passed to the base tokenizer's `encode_batch` method. While our code correctly handles the parameter after encoding, by that point the base tokenizer may have already added special tokens.

## Solution
This PR adds a check to see if the base tokenizer's `encode_batch` method supports the `add_special_tokens` parameter, and if so, passes it to ensure special tokens are not added by the base tokenizer. This provides backward compatibility with tokenizers that don't support this parameter.

Fixes #765

### DIFF
--- a/olmo/tokenizer.py
+++ b/olmo/tokenizer.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import inspect
 from pathlib import Path
 from typing import List, Optional, Union
 
@@ -180,7 +181,12 @@ class Tokenizer:
         if truncate_to is not None and add_special_tokens:
             truncate_to -= self.num_special_tokens_to_add(False)
 
-        batch_encoding = self.base_tokenizer.encode_batch(inputs)
+        # Check if the base tokenizer's encode_batch method supports add_special_tokens parameter
+        if 'add_special_tokens' in inspect.signature(self.base_tokenizer.encode_batch).parameters:
+            batch_encoding = self.base_tokenizer.encode_batch(inputs, add_special_tokens=False)
+        else:
+            # Fallback to original behavior if the parameter isn't supported
+            batch_encoding = self.base_tokenizer.encode_batch(inputs)
 
         all_input_ids = []
         for encoding in batch_encoding:


olmo/tokenizer.py
```diff
@@ -180,7 +181,12 @@ def encode_batch(self, inputs: List[str], add_special_tokens: bool = True) -> Li
         if truncate_to is not None and add_special_tokens:
             truncate_to -= self.num_special_tokens_to_add(False)
 
-        batch_encoding = self.base_tokenizer.encode_batch(inputs)
+        # Check if the base tokenizer's encode_batch method supports add_special_tokens parameter
+        if 'add_special_tokens' in inspect.signature(self.base_tokenizer.encode_batch).parameters:
+            batch_encoding = self.base_tokenizer.encode_batch(inputs, add_special_tokens=False)
```
dirkgr, line 186: This hard-codes `add_special_tokens` to `False`. Did you mean to pass it through from the arguments?

dirkgr (changes requested):

PR #831, opened 2025-04-25
Added .csv file which contains csv.gz files of training data.

aman-17:
1. Added .csv file which contains csv.gz files of training data.
2. Updated the README.


configs/official-1124/preprocessed_inputs.csv
```diff
@@ -0,0 +1,174 @@
+Dataset,Data Directory
```
liujch1998, line 1: Can we rename this file as `provenance.csv`? I think the .npy files have been called preprocessed inputs.

PR #850, opened 2025-06-11
Updated 1B Config

aman-17:
Updated seed42 config.yaml with respect to the over-written configs in training script: peteish1-anneal.sh.


aman-17:
Opened new PR and it is merged: #866

PR #881, opened 2025-08-28
support "upload" to local files

canonic-epicure:
support "upload" to local files


canonic-epicure:
Realized that remote upload just can be set to `null`. Nevermind.

